PR #306, opened 2023-04-12
fix: fix multi update map bug and polish db error

will-2012:
### Description

Fix multi update map bug and polish db error.

### Rationale

N/A.

### Example

N/A.

### Changes

Notable changes: 
* TaskNode/SQLDB.


sysvm (approved):
LGTM

PR #316, opened 2023-04-14
feat: update schema and order for list deleted objects

BarryTong65:
### Description

An additional response field `update_at` has been added, and data with the same `update_at` object will be sorted by object ID. 

### Rationale

N/A

### Example

N/A

### Changes

Notable changes: 
* Update new response field update_at
* Same `update_at` object will be sorted by object ID.


annielz (approved):

sysvm (approved):
LGTM

PR #319, opened 2023-04-15
feat: implement off-chain-auth solution

ruojunm:
### Description

For those dapp users, we here provide an off-chain-auth solution to access SP apis, so that  they don't have to make the signatures via the wallet popups over again and again.

### Changes

Notable changes: 
* add OffChainAuthKeyTable to store user's EDDSA public key
* add "request nonce" API to return the current nonce to the clients so that they can generate EDDSA key pairs based on the given nonce
* add "update user key" API to store user's EDDSA public key, so that later we can verify user's EDDSA sig 
* add "verify off chain auth sig" API and apply it in the request_uilt.go 's  verifySignature method
* add "verifyPersonalSignature" method in request_uilt.go, so that dapp users can use the wallet to sign the request to update their EDDSA public key
* add unit test


docs/run-book/01-deployment.md
```diff
@@ -59,6 +59,7 @@ Build   : go1.20.1 darwin amd64 2023-03-04 23:54
 ```shell
 # show services list
 ./gnfd-sp list
+auth                 off chain auth server
```
will-2012, line 62: Best to keep a consistent style, verbs that start with.

model/const.go
```diff
@@ -160,6 +165,10 @@ const (
 	// SignTypeV2 is an authentication algorithm, which is used by metamask
 	SignTypeV2 = "authTypeV2"
 
+	SignTypeOffChain   = "OffChainAuth" // sign type - off-chain-auth
```
will-2012, line 168: When to remove SignTypeV2?

store/mock/mockSPDB.go
```diff
@@ -0,0 +1,184 @@
+package mock
```
will-2012, line 1: `mockgen -source=store/sqldb/database.go -destination=store/sqldb/database_mock.go -package=sqldb` has existed in the makefile.

model/const.go
ruojunm: After Dcellar uses off-chain-auth, I will remove SignTypeV2.  So far, let's keep it and leave some time for Dcellar

constwz (approved):
LGTM

model/const.go
```diff
@@ -42,6 +44,7 @@ var SpServiceDesc = map[string]string{
 	MetadataService:    "Provides the ability to query meta data",
 	BlockSyncerService: "Syncs block data to db",
 	P2PService:         "Communicates with SPs on p2p protocol",
+	AuthService:        "off chain auth server",
```
annielz, line 47: Description start with upper case

2129zxl (approved):
LGTM

krish-nr (approved):
LGTM

BarryTong65:
LGTM

BarryTong65 (approved):

model/errors/rpc_error.go
```diff
@@ -98,6 +98,15 @@ var (
 	ErrInvalidAccountID = errors.New("invalid account id")
 )
 
+// auth service error
+var (
+	ErrInvalidRegNonce           = errors.New("invalid regNonce")
```
actioncli, line 103: add error description?

annielz (approved):

model/errors/rpc_error.go
ruojunm: Good catch. They are actually not used. I deleted all of them.

model/const.go
ruojunm: Good Catch. Fixed.

joeylichang (approved):

PR #330, opened 2023-04-18
feat: add task scheduler

joeylichang:
### Description

This PR adds scheduling task logic for the manager. the manager schedules the SP tasks, including uploading the object to the primary, replicating pieces to the secondary SP, and deleting the object.

The manager will consider two factors when assigning tasks to task nodes. The first is the priority of the task. The higher the priority, the greater the probability of being dispatched. Second, the number of resources available to the task node will not cause the task node to fail to execute tasks due to excessive resource consumption.

### Rationale

N/A

### Example

N/A

### Changes

Notable changes: 
* pkg/taskqueue
* service/manager

### DIFF
--- a/pkg/rcmgr/resource.go
+++ b/pkg/rcmgr/resource.go
@@ -11,13 +11,13 @@ import (
 // resources tracks the current state of resource consumption
 type resources struct {
 	limit        Limit
+	memory       int64
+	nfd          int
 	nconnsIn     int
 	nconnsOut    int
-	nfd          int
 	ntasksHigh   int
 	ntasksMedium int
 	ntasksLow    int
-	memory       int64
 }
 
 func addInt64WithOverflow(a int64, b int64) (c int64, ok bool) {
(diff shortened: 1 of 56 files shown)


pkg/rcmgr/resource.go
```diff
@@ -173,11 +176,92 @@ func (rc *resources) removeConns(incount, outcount, fdcount int) {
...
+	}
+	if low+rc.ntasksLow > rc.limit.GetTaskLimit(ReserveTaskPriorityLow) {
+		return &ErrTaskLimitExceeded{
+			current:   rc.ntasksLow,
+			attempted: low,
+			limit:     rc.limit.GetTaskLimit(ReserveTaskPriorityLow),
+			err:       fmt.Errorf("low priority task limit exceeded: %w", ErrResourceLimitExceeded),
+		}
+	}
+	rc.ntasksHigh += high
```
RenRick, line 222: Need to protect with atomic operation

will-2012: This component is not thread-safe, and the mutex is used when it is used externally. External users normally use resourceScope which has the mutex~

PR #336, opened 2023-04-19
feat: implement stop serving on testnet

forcodedancing:
### Description

This pr will implement the stop serving service for sp. With this service, sp can submit `MsgDiscontinueBucket` to blockchain to stop serving some buckets.

### Rationale

On testnet, we can limit the storage usage without reset.

### Example

NA

### Changes

Notable changes: 
* new service


service/stopserving/stopserving.go
```diff
@@ -0,0 +1,141 @@
...
+			s.discontinueBuckets()
+		case <-s.stopCh:
+			return
+		}
+	}
+}
+
+// discontinueBuckets fetch buckets from metadata service and submit transactions to chain
+func (s *StopServing) discontinueBuckets() {
+	createAt := time.Now().Unix() - int64(86400*s.config.DiscontinueConfig.BucketKeepAliveDays)
```
Fynnss, line 105: how about use `time.Now().AddDate(0, 0, -s.config.DiscontinueConfig.BucketKeepAliveDays)`?

service/stopserving/stopserving.go
```diff
@@ -0,0 +1,141 @@
...
+		}
+		txHash, err := s.signer.DiscontinueBucketOnChain(context.Background(), discontinueBucket)
+		if err != nil {
+			log.Errorw("failed to discontinue bucket on chain", "error", err)
+			return
+		} else {
+			s.cache.Add(bucket.BucketInfo.BucketName, struct{}{})
+			log.Infow("succeed to discontinue bucket", "bucket_name", discontinueBucket.BucketName, "tx_hash", txHash)
+		}
+		time.Sleep(SubmitTransactionInterval)
```
Fynnss, line 130: If `DiscontinueBucketOnChain ` uses the block mode to send transactions, still need sleep here?

service/stopserving/stopserving.go
```diff
@@ -0,0 +1,141 @@
...
+	createAt := time.Now().Unix() - int64(86400*s.config.DiscontinueConfig.BucketKeepAliveDays)
+	buckets, err := s.metadata.ListExpiredBucketsBySp(context.Background(), createAt, s.config.SpOperatorAddress, FetchBucketsLimit)
+	if err != nil {
+		log.Errorw("failed to query expired buckets", "error", err)
+		return
+	}
+
+	for _, bucket := range buckets {
+		bucketName := bucket.BucketInfo
+		if s.cache.Contains(bucketName) { // this bucket has been discontinued, however the metadata indexer did not handle it yet
```
Fynnss, line 114: Is it better to use BucketID ?

Fynnss: And For long-term thinking, we may use the same cache for discontinueBucket and discontinueObject. So, the key will has a prefix, E.g b{BucketID}, o{ObjectID}

service/stopserving/stopserving.go
forcodedancing: cool, fixed it

service/stopserving/stopserving.go
forcodedancing: yes, i agree.

service/stopserving/stopserving.go
forcodedancing: refactor to use the format you mentioned

will-2012:
LGTM. BTW, Has this part of the process been jointly tested with chain and metadata?

forcodedancing:
> LGTM. BTW, Has this part of the process been jointly tested with chain and metadata?

No joint test yet, need to do it on dev environment later.

service/stopserving/stopserving.go
```diff
@@ -0,0 +1,139 @@
...
+	createAt := time.Now().AddDate(0, 0, -s.config.DiscontinueConfig.BucketKeepAliveDays)
+	buckets, err := s.metadata.ListExpiredBucketsBySp(context.Background(), createAt.Unix(), s.config.SpOperatorAddress, FetchBucketsLimit)
+	if err != nil {
+		log.Errorw("failed to query expired buckets", "error", err)
+		return
+	}
+
+	for _, bucket := range buckets {
+		cacheKey := []byte{'b'}
+		cacheKey = append(cacheKey, bucket.BucketInfo.Id.Bytes()...)
```
Fynnss, line 112: It's better to use `GetBucketIDCacheKey` to wrapper this key format like we do it in greenfield repo

forcodedancing: Updated, https://github.com/bnb-chain/greenfield-storage-provider/pull/336/commits/dbdc59f286f47a0ff54cfddd03f867ba6f40d367

Fynnss (approved):
LGTM

will-2012 (approved):

PR #337, opened 2023-04-19
fix: permit anonymous users to access public object

will-2012:
### Description

1. Permit anonymous users to access the public object.
2. Refine some config to ensure to run action checker.

### Rationale

This should be allowed if the HTTP request has no `Authorization` header and gets the public object. And this change depends on the VerifyPermission interface behavior of the chain, and the chain needs to be upgraded at the same time.

Refer to: https://github.com/bnb-chain/greenfield/pull/186

### Example

N/A

### Changes

Notable changes: 
* Gateway.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -6,12 +6,14 @@ on:
       - master
       - develop
       - release*
+      - fix-release*
 
   pull_request:
     branches:
       - master
       - develop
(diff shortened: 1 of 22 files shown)


sysvm (approved):
LGTM

RenRick (approved):

RenRick:
LGTM

PR #340, opened 2023-04-19
feat: update metadata block syncer schema and add ListExpiredBucketsBySp method

annielz:
### Description
- Update block syncer schema according to changes on chain and events

- Add a new method ListExpiredBucketsBySp for SP internal GC usage 

### Rationale

This should suppport GC operation in SP

### Example

ListExpiredBucketsBySp(goCtx context.Context, req ListExpiredBucketsBySpRequest)(ListExpiredBucketsBySpResponse, error) 

type ListExpiredBucketsBySpRequest struct {
    CreateAt int64
    PrimarySpAddress string
    Limit int64
}

type ListExpiredBucketsBySpResponse struct {
	Buckets []*Bucket 
}

### Changes

Notable changes: 
* Gateway.
* Metadata Service.


store/bsdb/bucket.go
```diff
@@ -77,6 +77,22 @@ func (b *BsDBImpl) GetUserBucketsCount(accountID common.Address) (int64, error)
...
+// ListExpiredBucketsBySp lists expired buckets
+func (b *BsDBImpl) ListExpiredBucketsBySp(createAt int64, primarySpAddress string) ([]*Bucket, error) {
+	var (
+		buckets []*Bucket
+		err     error
+	)
+
+	err = b.db.Table((&Bucket{}).TableName()).
+		Select("*").
+		Where("primary_sp_address = ? and status = 'BUCKET_STATUS_CREATED' and create_time < ? and removed = false", common.HexToAddress(primarySpAddress), createAt).
```
clydemeng, line 93: 1.  Is it be possible to make the status as a parameter?  so that we can support more cases.
2. I remember you mentioned you will return the first 1000 rows. Do we still this behavior ?  How do we handle pagination ?

annielz: For 1, since bucket status only have 2 kinds: BUCKET_STATUS_CREATED, BUCKET_STATUS_DISCONTINUED, we only need BUCKET_STATUS_CREATED for the purpose of this specific query. If we will ever need to filter BUCKET_STATUS_DISCONTINUED status, "removed" in this query shall also become a parameter, and I feel the difference is large enough to be separate out as a stand alone new method.

For 2. returning first 1000 rows is the behavior of ListDeletedObjectsByBlockNumberRange. For this method, we will develop the pagination strategy along with GetUserBuckets in next phase.

will-2012:
Add a description based on the Description template, it is best not to modify the template format.

clydemeng (approved):

ruojunm (approved):

proto/service/metadata/types/metadata.proto
```diff
@@ -75,6 +101,14 @@ message GetUserBucketsCountRequest {
   string account_id = 1;
 }
 
+// ListExpiredBucketsBySpRequest is the request type for the ListExpiredBucketsBySp RPC method.
+message ListExpiredBucketsBySpRequest {
+  // create_at defines the create timestamp of bucket
+  int64 create_at = 1;
+  // primary_sp_address defines the primary sp address of bucket
+  string primary_sp_address = 2;
```
Fynnss, line 109: Don't support pagination?  If there are a large number of expired buckets, will returned at once?

proto/service/metadata/types/metadata.proto
```diff
@@ -25,6 +39,18 @@ message Object {
...
+  // delete_reason defines the deleted reason of object
+  string delete_reason = 6;
+  // operator defines the operator address of object
+  string operator = 7;
+  // create_tx_hash defines the creation transaction hash of object
+  string create_tx_hash = 8;
+  // update_tx_hash defines the update transaction hash of object
+  string update_tx_hash = 9;
+  // sealed_tx_hash defines the sealed transaction hash of object
+  string sealed_tx_hash = 10;
```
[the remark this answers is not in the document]
BarryTong65: fixed

proto/service/metadata/types/metadata.proto
forcodedancing: If no pagination, can we add another parameter `limit`, so we can limit the result and always get the oldest buckets for stop serving?

store/bsdb/bucket.go
annielz: Update: Pass in a new parameter for limit, can be within the range of 1-1000

proto/service/metadata/types/metadata.proto
annielz: Discussed and added a new parameter limit, which user can pass in 1 - 1000 as limit. Other limit input will be defaulted to 1000.

store/bsdb/bucket.go
```diff
@@ -82,15 +82,20 @@ func (b *BsDBImpl) GetUserBucketsCount(accountID common.Address) (int64, error)
...
 	)
 
+	if limit < 1 || limit > ExpiredBucketsDefaultSize {
+		limit = ExpiredBucketsDefaultSize
+	}
+
 	err = b.db.Table((&Bucket{}).TableName()).
 		Select("*").
 		Where("primary_sp_address = ? and status = 'BUCKET_STATUS_CREATED' and create_time < ? and removed = false", common.HexToAddress(primarySpAddress), createAt).
+		Limit(int(limit)).
```
forcodedancing, line 98: Do we need to put `Limit` after `Order`? In SQL we need to, I am not sure here.

service/metadata/client/metadata_client.go
```diff
@@ -104,6 +104,23 @@ func (client *MetadataClient) GetUserBucketsCount(ctx context.Context, in *metat
 	return resp, nil
 }
 
+// ListExpiredBucketsBySp list buckets that are expired by specific sp
+func (client *MetadataClient) ListExpiredBucketsBySp(ctx context.Context, createAt int64, primarySpAddress string, limit int64, opts ...grpc.CallOption) ([]*metatypes.Bucket, error) {
```
sysvm, line 108: the length of 108 line may exceed 120.

service/metadata/service/bucket.go
```diff
@@ -138,3 +172,41 @@ func (metadata *Metadata) GetUserBucketsCount(ctx context.Context, req *metatype
...
+				},
+				BucketStatus: types.BucketStatus(types.BucketStatus_value[bucket.Status]),
+			},
+			Removed:      bucket.Removed,
+			DeleteAt:     bucket.DeleteAt,
+			DeleteReason: bucket.DeleteReason,
+		})
+	}
+	resp = &metatypes.ListExpiredBucketsBySpResponse{Buckets: res}
+	log.CtxInfow(ctx, "succeed to get user buckets")
```
sysvm, line 210: fix `log.CtxInfow` to `log.CtxInfo`

service/metadata/client/metadata_client.go
annielz: This is to follow @joeylichang 's suggestion to pass in params other than grpc structs in client. Not sure if we can reduce length of a line while still keep its functionality?

joeylichang (approved):

proto/service/metadata/types/metadata.proto
Fynnss: BTW, I noticed that other API like `ListObjectsByBucketName` didn't support pagination or have a limit number. Could you add a todo to fix these. @annielz

PR #346, opened 2023-04-20
fix: add spdb and piece store metrics for downloader

sysvm:
### Description

Add spdb and piece store metrics for downloader.

### Rationale

N/A

### Example

N/A

### Changes

N/A


pkg/metrics/metric_items.go
```diff
@@ -45,4 +45,20 @@ var (
 		Name: "task_node_replicate_object_task_number",
 		Help: "Track task node service replicate object task",
 	}, []string{serviceLabelName})
+	// PieceStoreTimeHistogram records piece store request time
+	PieceStoreTimeHistogram = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "piece_store_handling_seconds",
+		Help:    "Track the latency for piece store requests",
+		Buckets: prometheus.DefBuckets,
```
will-2012, line 52: What is the meaning of `seconds` here?

service/downloader/downloader_service.go
```diff
@@ -76,6 +81,9 @@ func (downloader *Downloader) GetObject(req *types.GetObjectRequest,
 		log.CtxErrorw(ctx, "failed to check billing due to bucket quota", "error", err)
 		return merrors.InnerErrorToGRPCError(err)
 	}
+	observer := metrics.SPDBTimeHistogram.WithLabelValues("checkQuotaAndAddReadRecord")
+	observer.Observe(time.Since(startTime).Seconds())
```
will-2012, line 85: Is it possible to move to DB file?

pkg/metrics/metric_items.go
will-2012: The second level is not accurate enough?

sysvm: I use the same description as  https://github.com/grpc-ecosystem/go-grpc-middleware

service/downloader/downloader_service.go
sysvm: fixed

store/piecestore/client/piece_store_client.go
```diff
@@ -23,6 +25,11 @@ type StoreClient struct {
 	ps *piece.PieceStore
 }
 
+var (
```
will-2012, line 28: const?

sysvm: ok

PR #349, opened 2023-04-20
fix: keep retrying when any blocksycner event handles failure

krish-nr:
### Description

Fix blocksyncer events handle failure, if failed, blocksyncer will retired until success

### Rationale

As we cannot let any errors go when it happens when handling an event, so we catch every error when export events

### Example

N/A
### Changes

Notable changes: 
* add err handling when export and event failed

### DIFF
--- a/service/blocksyncer/block_syncer_indexer.go
+++ b/service/blocksyncer/block_syncer_indexer.go
@@ -44,15 +44,17 @@ func (i *Impl) ExportBlock(block *coretypes.ResultBlock, events *coretypes.Resul
 }
 
 // HandleEvent accepts the transaction and handles events contained inside the transaction.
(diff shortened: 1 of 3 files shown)


constwz (approved):

service/blocksyncer/block_syncer_indexer.go
```diff
@@ -86,7 +88,11 @@ func (i *Impl) Process(height uint64) error {
 	// 1. handle events in startBlock
 
 	if len(beginBlockEvents) > 0 {
-		i.ExportEventsWithoutTx(context.Background(), block, beginBlockEvents)
+		err = i.ExportEventsWithoutTx(context.Background(), block, beginBlockEvents)
+		if err != nil {
+			log.Errorf("failed to ExportEventsWithoutTx: %s", err)
```
sysvm, line 93: fix as `log.Errorf("failed to export events without tx: %s", err)`

sysvm (approved):
LGTM

PR #362, opened 2023-04-21
fix: sql paramthesis handling

annielz:
### Description

fix sql parenthesis handling for get object

### Rationale

for get object method to return correctly 

### Example

N/A

### Changes

Notable changes: 
* Metadata Service


krish-nr (approved):
LGTM

PR #384, opened 2023-04-26
feat: add universal endpoint view option 

annielz:
### Description

Add view option for universal endpoint and fix download file name without suffix

### Rationale

Users want to have a view option for universal endpoint, such that if they host html file on Dcellar, the universal endpoint will open the html in a new page other than download it as an attachment.
 
### Example

Universal Endpoint View link is like:
{endpoint}/view/{bucket_name}/{object_name}

### Changes

Notable changes: 
* Add new Universal Endpoint api in Downloader

### DIFF
--- a/model/const.go
+++ b/model/const.go
@@ -155,11 +155,11 @@ const (
 	ContentTypeJSONHeaderValue = "application/json"
 	// ContentTypeXMLHeaderValue is used to indicate xml
 	ContentTypeXMLHeaderValue = "application/xml"
-	//ContentDispositionHeader is used to indicate the media disposition of the resource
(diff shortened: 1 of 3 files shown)


model/const.go
```diff
@@ -159,6 +159,8 @@ const (
 	ContentDispositionHeader = "Content-Disposition"
 	//ContentDispositionAttachmentValue is used to indicate attachment
 	ContentDispositionAttachmentValue = "attachment"
+	//ContentDispositionInlineValue is used to indicate inline
```
will-2012, line 162: Add a space before commentary

ruojunm (approved):

will-2012 (approved):

PR #404, opened 2023-05-08
feat: list objects pagination & folder path

BarryTong65:
### Description

list objects pagination & folder path

### Rationale

N/A

### Example

```
curl --location --request GET 'http://barry.localhost:9033?continuation-token=&start-after&delimiter=/&prefix=my_folder1/&max-keys=9' \
--header 'Authorization: authTypeV2 ECDSA-secp256k1, Signature=1234567812345678123456781234567812345678123456781234567812345678'
```

### Changes

Notable changes: 
* pagination feature for Listobjectbybucketname
* add get user buckets default limit
* support folder path with prefix and delimiter query


service/gateway/metadata_handler.go
```diff
@@ -96,14 +111,75 @@ func (gateway *Gateway) listObjectsByBucketNameHandler(w http.ResponseWriter, r
...
 
-	if err = s3util.CheckValidBucketName(reqContext.bucketName); err != nil {
-		log.Errorw("failed to check bucket name", "bucket_name", reqContext.bucketName, "error", err)
+	queryParams = reqContext.request.URL.Query()
+	requestBucketName = reqContext.bucketName
+	requestMaxKeys = queryParams.Get("max_keys")
+	requestStartAfter = queryParams.Get("start_after")
+	requestContinuationToken = queryParams.Get("continuation_token")
+	requestDelimiter = queryParams.Get("delimiter")
+	requestPrefix = queryParams.Get("prefix")
```
will-2012, line 120: The constant queries field related to restful api can be defined in model const.go

service/gateway/param_util.go
```diff
@@ -0,0 +1,40 @@
...
+		}
+	}
+	return false
+}
+
+// isValidObjectPrefix checks if the given object prefix is valid:
+// - does not have invalid path segments
+// - is a valid UTF-8 string
+// - does not contain double slashes "//"
+func isValidObjectPrefix(prefix string) bool {
```
will-2012, line 29: checkValidObjectPrefix?

service/metadata/service/object.go
```diff
@@ -2,59 +2,107 @@ package service
...
+	}
+
+	// returns some or all (up to 1000) of the objects in a bucket with each request
+	if req.MaxKeys > model.ListObjectsLimitSize {
+		maxKeys = model.ListObjectsLimitSize
+	}
 
-	objects, err := metadata.bsDB.ListObjectsByBucketName(req.BucketName)
+	ctx = log.Context(ctx, req)
+	results, err = metadata.bsDB.ListObjectsByBucketName(req.BucketName, req.ContinuationToken, req.Prefix, req.Delimiter, int(maxKeys))
```
will-2012, line 40: req. StartAfter is unused?

BarryTong65: The "start_after" is used to specify the starting point of the query, while "next_continuation_token" is used to persist the query state for use in the next query. These two are not used simultaneously.

Within the code logic, if the next_continuation_token is empty, the "start_after" will assign the value to the marker (ContinuationToken). This approach is based on the logic found in the ListObjectV2 function of Minio (similar to AWS S3).

service/gateway/param_util.go
BarryTong65: fixed

service/gateway/metadata_handler.go
BarryTong65: fixed

service/metadata/service/object.go
will-2012: If this field is not used, maybe it can be removed in the proto definition.

model/const.go
```diff
@@ -200,6 +200,16 @@ const (
 	ListBucketReadRecordQuery = "list-read-record"
 	// ListBucketReadRecordMaxRecordsQuery defines list read record max num
 	ListBucketReadRecordMaxRecordsQuery = "max-records"
+	// ListObjectsMaxKeysQuery defines the maximum number of keys returned to the response
+	ListObjectsMaxKeysQuery = "max_keys"
```
will-2012, line 204: Refine `max_keys` to `max-keys`?

model/const.go
```diff
@@ -200,6 +200,16 @@ const (
...
+	// ListObjectsStartAfterQuery defines where you want to start listing from
+	ListObjectsStartAfterQuery = "start_after"
```
will-2012, line 206: Ditto.

model/const.go
```diff
@@ -200,6 +200,16 @@ const (
...
+	// ListObjectsContinuationTokenQuery indicates that the list is being continued on this bucket with a token
+	ListObjectsContinuationTokenQuery = "continuation_token"
```
will-2012, line 208: Ditto.

store/bsdb/object.go
```diff
@@ -1,18 +1,61 @@
...
+				LIMIT ?;`,
+			bucketName, prefix+"%", continuationToken, continuationToken, delimiter, prefix,
+			prefix, delimiter, prefix, delimiter,
+			bucketName, prefix+"%", continuationToken, continuationToken, delimiter, prefix,
+			limit).Scan(&results).Error
+	} else {
+		// If delimiter is not specified, retrieve objects directly
+		err = b.db.Table((&Object{}).TableName()).
+			Select("*").
+			Where("bucket_name = ? AND object_name >= IF(? = '', '', ?)", bucketName, continuationToken, continuationToken).
```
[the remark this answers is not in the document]
BarryTong65: image
image
image
image
When the prefix and continuationToken are empty in a SQL query, there is still a difference in the query time and response, which can result in additional overhead. To handle this situation of empty strings, I added a filter.

```
if continuationToken != "" {
	filters = append(filters, ContinuationTokenFilter(continuationToken))
}
if prefix != "" {
	filters = append(filters, PrefixFilter(prefix))
}
```

service/metadata/service/object.go
BarryTong65: "start_after" is being used and it is assigned a value to the data at the handler level, the "start_after" will assign the value to the marker (ContinuationToken).

```
if requestContinuationToken == "" {
            continuationToken = requestStartAfter
}
```
https://github.com/minio/minio/blob/b92cdea5789aab0e71f98c7dbdca9d5c8d22a6fe/cmd/erasure-server-pool.go#L1191

model/const.go
BarryTong65: fixed

model/const.go
BarryTong65: fixed

model/const.go
BarryTong65: fixed

will-2012 (approved):

ruojunm (approved):

PR #407, opened 2023-05-08
feat: add full-memory replicate task

will-2012:
### Description

Add full-memory replicate task.

### Rationale

There are currently two replicate task implementations, full memory and streaming. Streaming takes up less memory, but in complex network conditions, the transmission is slower. The specific reason needs to be further located and optimized in the future. Full memory takes up more memory, but the transmission is faster, and full memory is used by default.

### Example

N/A.

### Changes

Notable changes: 
* p2p: reduce debug log.
* gateway: fix invalid header.
* tasknode: add full-mem replicate task.

### DIFF
--- a/pkg/p2p/node.go
+++ b/pkg/p2p/node.go
@@ -37,12 +37,12 @@ type Node struct {
 	stopCh            chan struct{}
 }
 
-// NewNode return an instance of Node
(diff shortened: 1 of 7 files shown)


joeylichang (approved):

sysvm (approved):
LGTM

PR #462, opened 2023-05-24
fix: base app rcmgr span panic

joeylichang:
### Description

Fix the bug that approver and downloader misuse the receiver rcmgr, and add protection for nil module.

### Rationale

N/A

### Example

N/A

### Changes

Notable changes: 

N/A


sysvm (approved):
LGTM

will-2012 (approved):

PR #468, opened 2023-05-24
chore: add error for cal nil model

joeylichang:
### Description

 Add error for cal nil model, some config error may call nil model, if return nil, it is diffcult to debug.

### Rationale

N/A

### Example

N/A

### Changes

Notable changes: 

N/A


sysvm (approved):
LGTM

will-2012 (approved):

PR #528, opened 2023-06-01
feat: support loading tasks

will-2012:
### Description

Manager support loading tasks at startup.

### Rationale

The manager can load pending tasks from sp db, and then the tasks will be executed in the executor.
Currently include replicate/seal/gc-object task.

### Example

N/A.

### Changes

Notable changes: 
* sp-db: refactor sp db, remove useless interfaces.
* manager: add load impl at startup.


sysvm (approved):
LGTM

base/gfspconfig/config.go
```diff
@@ -222,3 +223,7 @@ type MetadataConfig struct {
 	IsMasterDB                 bool
 	BsDBSwitchCheckIntervalSec int64
 }
+
+type ManagerConfig struct {
```
joeylichang, line 227: load should be a must, why need config?

modular/manager/manager.go
```diff
@@ -42,6 +44,12 @@ type ManageModular struct {
 	scope   rcmgr.ResourceScope
 	mux     sync.Mutex
 
+	// loading task at startup.
+	enableLoadTask           bool
+	loadTaskLimitToReplicate int
```
joeylichang, line 49: limit should use the queue cap.

base/gfspconfig/config.go
will-2012: If there is a bug in the load, it may not be able to start; this may not be acceptable, and the configuration is more flexible.

modular/manager/manager.go
will-2012: Yes, can't get cfg here and `manager.loadTaskLimitToReplicate = cfg.Parallel.GlobalReplicatePieceParallel`.

PR #538, opened 2023-06-03
feat: support loading tasks v0.2.1

will-2012:
### Description

Manager support loading tasks at startup, which is cherry-picked from https://github.com/bnb-chain/greenfield-storage-provider/pull/528.

### Rationale

The manager can load pending tasks from sp db, and then the tasks will be executed in the executor.
Currently include replicate/seal/gc-object task.

### Example

N/A.

### Changes

Notable changes: 
* sp-db: refactor sp db, remove useless interfaces.
* manager: add load impl at startup.


sysvm (approved):
LGTM

will-2012:
This PR is not submitted to 0.2.1.

PR #542, opened 2023-06-05
feat: change get block height by ws protocol

joeylichang:
### Description

Change get block height by ws protocol, it is more lightweight.

### Rationale

N/A

### Example

N/A

### Changes

Notable changes: 

N/A


sysvm (approved):
LGTM

PR #547, opened 2023-06-06
feat: add removed param for metadata service apis

annielz:
### Description

Adding "removed" param for metadata service apis 

### Rationale
1. For metadata service apis returning single object, we default removed = false into the db query 
2. For metadata service apis returning a list, we add input param "removed" for users to filter the list

### Example

N/A

### Changes

Notable changes: 
* metadata service


ruojunm (approved):

BarryTong65:
what is the difference between includeRemoved and includePrivate?

PR #614, opened 2023-06-16
fix: modify the default gc opt to a suitable value v0.2.2

will-2012:
### Description

Modify the default gc opt to a suitable value which is cherry-picked from https://github.com/bnb-chain/greenfield-storage-provider/pull/613.

### Rationale

Previously, a gc task of 500 blocks was generated every half hour, but the blockchain generated 900 blocks every half hour; resulting in the gc task not being able to keep up with the latest high speed.

### Example

N/A.

### Changes

Notable changes: 
* Manager default opt.


sysvm (approved):
LGTM

joeylichang (approved):

PR #622, opened 2023-06-19
fix: fix missing gc object v0.2.2

will-2012:
### Description

Fix missing gc object, which is cherry-picked from https://github.com/bnb-chain/greenfield-storage-provider/pull/621.

### Rationale

When objects are in different blocks, object ids are unordered. The check order may cause some objects not to be gc.

### Example

N/A.

### Changes

Notable changes: 
* GC object workflow, fix check conditions.


actioncli (approved):

joeylichang (approved):

sysvm (approved):
LGTM

PR #629, opened 2023-06-19
fix : fix compute ec piece size and add task verify 

flywukong:
### Description

1. fix ECPieceSize, the original way ignoring the padding
2. replicate handler missed verifying the signature of receive task , add it

### Rationale

fix code issue

### Example

### Changes

Notable changes: 
* ...


sysvm:
Please check e2e action.

PR #696, opened 2023-06-29
feat:  blocksyncer adapt latest changes for sp meta from chain side( with juno version updated)

krish-nr:
### Description

blocksyncer adapt latest changes for sp meta from chain side

### Rationale

sp data structure chaged(replace operatore address with spId), we adapted this change in juno and this PR is to update dependency
 
### Example

N/A

### Changes

Notable changes: 
*  juno version updated


sysvm (approved):
LGTM

PR #699, opened 2023-06-30
feat: SP database sharding

annielz:
### Description

Sharding for SP database tables which include:
BSDB: object, slash_prefix_tree_nodes
SPDB: object_integrity

Each of the above tables will be sharded to 64 tables.

### Rationale

We want to optimize the query speed with no breaking change, before the Mainnet get launched. After the Mainnet launch, the object size will soon exceed 5,000,000 lines, which would degrade Mysql performance.

### Example

None of the API usage shall feel any difference. The changes are made in the db level.

### Changes

Notable changes: 
* bsdb
* spdb
* metadata service

### DIFF
--- a/store/bsdb/object.go
+++ b/store/bsdb/object.go
@@ -49,7 +49,7 @@ func (b *BsDBImpl) ListObjectsByBucketName(bucketName, continuationToken, prefix
 		}
 
 		if includeRemoved {
-			err = b.db.Scopes(ReadObjectsTable(bucketName)).
+			err = b.db.Table(GetObjectsTableName(bucketName)).
 				Select("*").
 				Where("bucket_name = ?", bucketName).
 				Scopes(filters...).
@@ -57,7 +57,7 @@ func (b *BsDBImpl) ListObjectsByBucketName(bucketName, continuationToken, prefix
 				Order("object_name asc").
 				Find(&results).Error
 		} else {
-			err = b.db.Scopes(ReadObjectsTable(bucketName)).
+			err = b.db.Table(GetObjectsTableName(bucketName)).
 				Select("*").
 				Where("bucket_name = ? and removed = false", bucketName).
 				Scopes(filters...).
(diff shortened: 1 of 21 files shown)


flywukong (commented):
do we have any compare performance test data of database sharding ? It is better to provide some performance comparison data

annielz:
> do we have any compare performance test data of database sharding ? It is better to provide some performance comparison data

Screen Shot 2023-07-06 at 09 35 33

Screen Shot 2023-07-06 at 09 35 54

When testing on QA environment, which currently have 1,337,708 lines in object table, querying one object metadata from an unsharded DB table of sp-f namespace costs 263ms, while queuing one object metadata from an sharded DB table of sp-h namespace costs 253ms, a little faster than unsharded DB table.  

Generally,  it is a best practice to keep a table within 5,000,000 lines. The performance difference will become more obvious when the data exceed 20,000,000 lines per table in unsharded DB table(more than 3 layers for mysql b+ tree structure).

store/bsdb/object.go
```diff
@@ -128,11 +166,34 @@ func (b *BsDBImpl) ListObjectsByObjectID(ids []common.Hash, includeRemoved bool)
 	if !includeRemoved {
 		filters = append(filters, RemovedFilter(includeRemoved))
 	}
+	for _, id := range ids {
+		var object *Object
+		bucketName, err := b.GetBucketNameByObjectID(id)
+		if err != nil {
+			continue
```
BarryTong65, line 173: I think err here should be recorded in log

store/bsdb/object.go
```diff
@@ -128,11 +166,34 @@ func (b *BsDBImpl) ListObjectsByObjectID(ids []common.Hash, includeRemoved bool)
...
+		}
+		err = b.db.Table(GetObjectsTableName(bucketName)).
+			Select("*").
+			Where("object_id = ?", id).
+			Scopes(filters...).
+			Take(&object).Error
+		if err != nil {
+			continue
```
BarryTong65, line 181: ditto

store/bsdb/object.go
annielz: Good point, just added error log

store/bsdb/object.go
annielz: Also just added error log here

BarryTong65 (approved):

clydemeng (approved):

ruojunm (approved):

PR #706, opened 2023-06-30
feat: verify object permission by meta service

irrun:
### Description

feat: verify object permission by meta service

### Rationale

### Example

### Changes

Notable changes: 
* ...


BarryTong65 (approved):

sysvm (approved):
LGTM

joeylichang:
1. Will it support permission verification for PutObject and other operations in the future?
2. The metadata permission check is mainly to query the db. Is it possible to put the logic of the metadata permission check into the authenticator, which can reduce one RPC request.

PR #728, opened 2023-07-04
feat: add SAIAMType for oss newAliyunfsSession

augustyip:
### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...


store/piecestore/storage/aliyunfs.go
```diff
@@ -2,11 +2,14 @@ package storage
 
 import (
 	"fmt"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
```
sysvm, line 6: Imported pkg should be sorted, order: std lib, third-party lib, inner lib.

PR #733, opened 2023-07-05
fix: fix challenge recovery

flywukong:
### Description
remove challenge recovery condition.  If calling PieceStore().GetPiece got an error in the download task,  it is the ErrPieceStore, no need to judge the inner code.

### Rationale

inner code error is not supported  in download task when calling PieceStore().GetPiece

### Example

NA

### Changes

Notable changes: 
* remove challenge recovery condition

### DIFF
--- a/modular/downloader/download_task.go
+++ b/modular/downloader/download_task.go
@@ -3,9 +3,7 @@ package downloader
 import (
 	"context"
 	"errors"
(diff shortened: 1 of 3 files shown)


joeylichang (approved):

sysvm (approved):
LGTM

PR #752, opened 2023-07-07
fix: remove the private_key from the code

ruojunm:
### Description

remove the private_key from the code
### Rationale

We should not keep the private key in the code , even it is for test purpose

### Changes

Notable changes: 
* generate the private key by using ether js package
* put the generated private key and account address into the e2e test action setp

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -35,6 +35,22 @@ jobs:
       - name: Run MySQL
         run: |
           sudo /etc/init.d/mysql start
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14.x'
+      - name: Install dependencies
+        run: npm install ethers
+      # generate test private key
+      - name: Generate private key
+        run: node ./.github/workflows/generate-private-key.js
+        id: generate-private-key
+      - name: set private key and account address as job variables
+        run: |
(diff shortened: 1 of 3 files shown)


sysvm (approved):
LGTM

.github/workflows/e2e-test.yml
```diff
@@ -35,6 +35,22 @@ jobs:
...
+        run: npm install ethers
+      # generate test private key
+      - name: Generate private key
+        run: node ./.github/workflows/generate-private-key.js
+        id: generate-private-key
+      - name: set private key and account address as job variables
+        run: |
+          echo "Private key: ${{ steps.generate-private-key.outputs.private_key }}"
+          echo "Account address: ${{ steps.generate-private-key.outputs.account_address }}"
+          echo "PRIVATE_KEY=${{ steps.generate-private-key.outputs.private_key }}" >> "$GITHUB_ENV"
```
sysvm, line 52: what does GITHUB_ENV mean? I don't see the use of GITHUB_ENV.

will-2012 (approved):

.github/workflows/e2e-test.yml
ruojunm: It is used to pass values between steps:

https://docs.github.com/en/actions/learn-github-actions/variables#passing-values-between-steps-and-jobs-in-a-workflow

PR #753, opened 2023-07-07
fix: add verify receive task signature and verify primary SP

flywukong:
### Description

the receiver needs to check the signature of the primary SP and check the request account if it is really the primary on chain

### Rationale

fix receiver

### Example

NA

### Changes

Notable changes: 
*  fix receiver handler


sysvm (approved):
LGTM

joeylichang (approved):

PR #758, opened 2023-07-07
feat: refactor recovery and support recovery secondary from other secondary SPs

flywukong:
### Descriptions
1. if the recovery task is to recover secondary SP data, 
（1） first, it sends a request to the primary SP to get the corresponding piece and recover the data   
（2）if it fails to recover from the primary SP, it needs to send requests to other secondary SPs,  use the pieces to decode the segment and recover the piece of it

2. remove the code of download recovery  which make download and challenge recovery not be triggered automatically

3. remove the get-piece interface which offers SDK piece data to recovery

4. support "recovery.piece" command to recover a single segment of the primary SP or secondary SP

### Rationale
need to support recovery secondary from other secondary SPs

### Example
NA

### Changes

Notable changes: 
* support recovery secondary from other secondary SPs


modular/executor/execute_task.go
```diff
@@ -337,143 +335,247 @@ func (e *ExecuteModular) HandleRecoverPieceTask(ctx context.Context, task coreta
...
+		int(dataShards),
+		int(parityShards))
+	if err != nil {
+		log.CtxErrorw(ctx, "failed to ec encode data when recovering secondary SP", "error", err)
+		return nil, err
+	}
+	return ECEncodeData[redundancyIdx], nil
 }
 
 func (e *ExecuteModular) checkRecoveryCheckSum(ctx context.Context, task coretask.RecoveryPieceTask, recoveryChecksum []byte) error {
```
sysvm, line 568: fix `checkRecoveryCheckSum` as `checkRecoveryChecksum`

sysvm (approved):
LGTM

chris-Li-Node-Real (approved):

PR #776, opened 2023-07-10
fix: code security check

joeylichang:
### Description

* Accept
   - [x] Sensitive data returned by HTTP request headers flows to log
   - [x] Prefer SHA256 or stronger algorithm over MD5
   - [x] Potential Slowloris Attack because ReadHeaderTimeout is not configured in the http.Server
   - [x] Deferring unsafe method "Close" on type "*os.File"
   - [x] Error not handled(UpdateJobState)
   - [x] MinVersion is missing from this TLS configuration
   - [x] Non deterministic select
* Acknowledge
  - [x] Incorrect conversion of an integer
           strconv.Atoi includes upper bound check
   - [x] Prefer crypto/rand instead of math/rand
            rand is not involved in the security of user data and user content   
   - [x] Found an insecure gRPC server without 'grpc.Creds()' or options with credentials 
           GRPC is only used between internal services, no user requests

### Rationale

N/A

### Example

N/A

### Changes

Notable changes: 

N/A

### DIFF
--- a/base/gfspclient/client.go
+++ b/base/gfspclient/client.go
@@ -2,6 +2,7 @@ package gfspclient
 
 import (
 	"context"
+	"crypto/tls"
 	"net/http"
 	"sync"
 	"time"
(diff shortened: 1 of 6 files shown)


sysvm (approved):
LGTM

will-2012 (approved):

PR #809, opened 2023-07-13
fix: remove checking account if it is exists in greenfield

flywukong:
### Description

The logic of checking the existence of the account has been removed from the VerifyAuthentication function which is used to verify authorization of the request to SP.  The reasons for the removal are as follows:：
1）Data marketplace users only operate on BSC and have no transactions on Greenfield.
2）During batch upload, a temporary account is generated, and the gas and fees for this temporary account are deducted from the user's account (with authorization). This temporary account has not yet been created on the chain, but actions like createObject are performed

### Rationale

support marketplace user case

### Example

NA

### Changes

Notable changes: 
*  remove checking the existence of the account

### DIFF
--- a/modular/authenticator/authenticator.go
+++ b/modular/authenticator/authenticator.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	storagetypes "github.com/bnb-chain/greenfield/x/storage/types"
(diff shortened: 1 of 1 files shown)


sysvm (approved):
LGTM

PR #842, opened 2023-07-17
fix: integrity hash not being updated in resumable upload

jingjunLi:
### Description

Fix the issue of integrity hash not being updated in resumable upload.

### Rationale

* add `UpdatePieceChecksum`
1) If the IntegrityMetaTable does not exist, it will be created.
2) If the IntegrityMetaTable already exists, it will be appended to the existing PieceChecksumList.
* add `UpdateIntegrityChecksum`
update integrity hash info to db

### Example

NA

### Changes

Notable changes: 
* fix the case where Passwd and Address are both empty in DefaultGfSpDBOption func.
* ...

### DIFF
--- a/base/gfspapp/app_options.go
+++ b/base/gfspapp/app_options.go
@@ -255,10 +255,10 @@ func DefaultGfSpDBOption(app *GfSpBaseApp, cfg *gfspconfig.GfSpConfig) error {
 		cfg.SpDB.User = "root"
 	}
 	if cfg.SpDB.Passwd == "" {
-		cfg.SpDB.User = "test"
+		cfg.SpDB.Passwd = "test"
 	}
 	if cfg.SpDB.Address == "" {
-		cfg.SpDB.User = "127.0.0.1:3306"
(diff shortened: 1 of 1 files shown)


store/sqldb/object_integrity.go
```diff
@@ -202,15 +235,14 @@ func (s *SpDBImpl) AppendObjectChecksumIntegrity(objectID uint64, redundancyInde
 	} else {
 		newChecksums := append(integrityMeta.PieceChecksumList, checksum)
 		integrityMeta.PieceChecksumList = newChecksums
-		result := s.db.Model(&IntegrityMetaTable{}).Where("object_id = ? and redundancy_index = ?", objectID, redundancyIndex).
+		result := s.db.Table(shardTableName).Where("object_id = ? and redundancy_index = ?", objectID, redundancyIndex).
```
sysvm, line 238: This case can use `UpdateIntegrityChecksum` function?

sysvm: OK

sysvm (approved):
LGTM

PR #850, opened 2023-07-18
fix: handle concurrent spdb table creation

annielz:
### Description

As there are scenarios where multiple pods start together, it is possible when we cleanup spdb and recreate tables with autoMigrate, there will be race condition when creating a table which another pod just created, thus throw an error. We want to silent such "table already exists" error to not throw and return nil db pointer, but continue execution onto the next table.

### Rationale

spdb init db logic update.

### Example

N/A

### Changes

Notable changes: 
* spdb


sysvm (approved):
LGTM

BarryTong65 (approved):

PR #877, opened 2023-07-24
feat: reduce ci

jingjunLi:
### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...


will-2012 (approved):

will-2012 (changes requested):
Update dest branch to develop.

PR #907, opened 2023-07-26
feat: pick sp gvg not exiting

jingjunLi:
### Description

pick sp gvg not exiting.

### Rationale

During the bucket migration process, we have a source Global Virtual Group (GVG) with a list of members [3, 1, 2, 4, 5, 6, 7]. The goal is to organize the destination GVG by replacing any member from the source GVG that is in STATUS_GRACEFUL_EXITING.
To achieve this, we compare the members of the source GVG with the destination GVG. If we find a member from the source GVG that is in STATUS_GRACEFUL_EXITING, we replace it in the destination GVG. 

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/modular/manager/bucket_migrate_scheduler.go
+++ b/modular/manager/bucket_migrate_scheduler.go
@@ -438,6 +438,32 @@ func (s *BucketMigrateScheduler) createGlobalVirtualGroupForBucketMigrate(vgfID 
 	})
 }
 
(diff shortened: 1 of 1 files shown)


Fynnss:
> Global Verifiable Group

 what is the Global Verifiable Group?

will-2012 (approved):

Fynnss (approved):

PR #910, opened 2023-07-27
fix: return err when request height > latest

BarryTong65:
### Description

return err when request block id is larger than latest block height for  swapout/spexit/mirgration bucket event apis

### Rationale

N/A
### Example

N/A

### Changes

Notable changes: 
* return err when request block id is larger than latest block height for  swapout/spexit/mirgration bucket event apis


will-2012 (approved):

sysvm (approved):
LGTM

PR #924, opened 2023-07-28
fix: refine sp exit scheduler

will-2012:
### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/modular/manager/sp_exit_scheduler.go
+++ b/modular/manager/sp_exit_scheduler.go
@@ -297,14 +297,7 @@ func (s *SPExitScheduler) AddSwapOutToTaskRunner(swapOut *virtualgrouptypes.MsgS
 	}
 
 	for _, gvg := range gvgList {
-		redundancyIndex := int32(-1)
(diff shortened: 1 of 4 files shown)


modular/manager/sp_exit_scheduler.go
```diff
@@ -316,6 +309,7 @@ func (s *SPExitScheduler) AddSwapOutToTaskRunner(swapOut *virtualgrouptypes.MsgS
 			log.Errorw("failed to add swap out to task runner", "gvg_unit", gUnit, "error", err)
 			return err
 		}
+		log.Infow("succeed to add gvg unit to task runner", "gvg_unit", gUnit)
```
Fynnss, line 312: Also need to print the info of sp and gvg.

will-2012: gUnit already contains all.

sysvm (approved):
LGTM

Fynnss (approved):

PR #966, opened 2023-08-02
fix: refine migrate gvg task workflow

will-2012:
### Description

Refine migrate gvg task workflow.

### Rationale

1. Infinitely retry migration tasks in case of failure, only be gc when it is finished.
2. When the task is not updated within a certain period of time, it will be executed by other executors.

### Example

N/A.

### Changes

Notable changes: 
* Refine migrate gvg task retry policy and gc policy.
* Refine some log and add metrics.

### DIFF
--- a/base/gfspapp/task_options.go
+++ b/base/gfspapp/task_options.go
@@ -52,9 +52,9 @@ const (
 	// MaxMigratePieceTime defines the max timeout to migrate piece.
 	MaxMigratePieceTime int64 = 50
 	// MinMigrateGVGTime defines the min timeout to migrate gvg.
-	MinMigrateGVGTime int64 = 10
+	MinMigrateGVGTime int64 = 1800 // 0.5 hour
 	// MaxMigrateGVGTime defines the max timeout to migrate gvg.
-	MaxMigrateGVGTime int64 = 50
+	MaxMigrateGVGTime int64 = 3600 // 1 hour
 
 	// NotUseRetry defines the default task max retry.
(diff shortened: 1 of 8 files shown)


sysvm (approved):
LGTM

base/gfspapp/task_options.go
```diff
@@ -52,9 +52,9 @@ const (
 	// MaxMigratePieceTime defines the max timeout to migrate piece.
 	MaxMigratePieceTime int64 = 50
 	// MinMigrateGVGTime defines the min timeout to migrate gvg.
-	MinMigrateGVGTime int64 = 10
+	MinMigrateGVGTime int64 = 1800 // 0.5 hour
 	// MaxMigrateGVGTime defines the max timeout to migrate gvg.
-	MaxMigrateGVGTime int64 = 50
+	MaxMigrateGVGTime int64 = 3600 // 1 hour
```
Fynnss, line 57: Is the 1 hour timeout also for GVGs with large data storage?

will-2012: if the task is running, the update time will be updated, and will not be judged as timed out ( update_time + this var > now).

Fynnss: In other words, will the start time of the timeout be updated?

Fynnss: > if the task is running, the update time will be updated, and will not be judged as timed out ( update_time + this var > now).

Okay, make sense.

Fynnss (approved):
LGTM

PR #973, opened 2023-08-03
fix: add missing path for pprof server

j75689:
### Description

- allocs      : "A sampling of all past memory allocations"
- block       : "Stack traces that led to blocking on synchronization primitives"
- goroutine   : "Stack traces of all current goroutines. Use debug=2 as a query parameter to export in the same format as an unrecovered panic."
- heap        : "A sampling of memory allocations of live objects. You can specify the gc GET parameter to run GC before taking the heap sample."
- mutex       : "Stack traces of holders of contended mutexes"
- threadcreate: "Stack traces that led to the creation of new OS threads"

### Rationale

### Example

### Changes

Notable changes: 
* ...


xibao-nr (approved):

will-2012 (approved):

sysvm (approved):
LGTM

Fynnss (approved):

PR #975, opened 2023-08-04
fix:release v0.2.3 bs bug

constwz:
### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...


ruojunm (approved):

modular/blocksyncer/blocksyncer_options.go
```diff
@@ -187,15 +187,17 @@ func (b *BlockSyncerModular) serve(ctx context.Context) {
 
 // enqueueNewBlocks enqueues new block heights onto the provided queue.
 func (b *BlockSyncerModular) enqueueNewBlocks(context context.Context, exportQueue types.HeightQueue, currHeight uint64) {
+	ticker := time.NewTicker(1500 * time.Millisecond)
 	// Enqueue upcoming heights
```
KeefeL, line 191: defer ticker.Stop()

modular/blocksyncer/blocksyncer_options.go
```diff
@@ -187,15 +187,17 @@ func (b *BlockSyncerModular) serve(ctx context.Context) {
...
 	for {
 		select {
 		case <-context.Done():
 			log.Infof("Receive cancel signal, enqueueNewBlocks routine will stop")
 			// close channel
 			close(exportQueue)
+			ticker.Stop()
```
KeefeL, line 198: No need here, please use defer ticker.Stop()

modular/blocksyncer/blocksyncer_options.go
```diff
@@ -187,6 +187,8 @@ func (b *BlockSyncerModular) serve(ctx context.Context) {
 
 // enqueueNewBlocks enqueues new block heights onto the provided queue.
 func (b *BlockSyncerModular) enqueueNewBlocks(context context.Context, exportQueue types.HeightQueue, currHeight uint64) {
+	ticker := time.NewTicker(1500 * time.Millisecond)
```
KeefeL, line 190: Suggest 1500ms -> 500ms, the block interval is around 2s, if we use 1.5ms, that means in some case, you should wait around 1s to get the latest block, it's too slow.

modular/blocksyncer/blocksyncer_options.go
```diff
@@ -206,17 +208,18 @@ func (b *BlockSyncerModular) enqueueNewBlocks(context context.Context, exportQue
 				}
 			}
 		}
-		time.Sleep(time.Second)
 	}
 }
 
 func (b *BlockSyncerModular) getLatestBlockHeight(ctx context.Context) {
+	ticker := time.NewTicker(1500 * time.Millisecond)
```
KeefeL, line 215: The same here, 500ms is ok.

modular/blocksyncer/blocksyncer_options.go
constwz: This coroutine is to get the latest height to write to the queue, may write several blocks of height in one request, and then another coroutine to consume, consumption takes time, so we do not always need to get the latest height

KeefeL (approved):

modular/blocksyncer/blocksyncer_options.go
```diff
@@ -286,17 +298,20 @@ func (b *BlockSyncerModular) fetchData(start, end uint64) {
...
 					continue
 				}
+				metrics.ChainRPCTime.Set(float64(time.Since(rpcStartTime).Milliseconds()))
 
 				events, err := b.parserCtx.Node.BlockResults(int64(height))
 				if err != nil {
 					log.Warnf("failed to get block results from node: %s", err)
 					continue
 				}
+				metrics.ChainRPCTime.Set(float64(time.Since(rpcStartTime).Milliseconds()))
```
ruojunm, line 314: for this metrics, do we need an new rpcStartTime?

ruojunm (approved):

PR #978, opened 2023-08-04
[NO NEED To MERGE] this pr is created just for reviewing the blocksyncer

guagualvcha:
### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...


modular/blocksyncer/blocksyncer_options.go
```diff
@@ -1,478 +0,0 @@
...
-				return err
-			}
-		}
-	}
-	return nil
-}
-
-// serve start BlockSyncer rpc service
-func (b *BlockSyncerModular) serve(ctx context.Context) {
-	migrateDBAny := ctx.Value(MigrateDBKey{})
```
guagualvcha, line 146: To me, the `MigrateDBKey` seems always true, do we really need this?

modular/blocksyncer/blocksyncer_options.go
```diff
@@ -1,478 +0,0 @@
...
-}
-
-func StopMainService() error {
-	CancelMain()
-	return nil
-}
-
-// mustGetLatestHeight tries getting the latest height from the RPC client.
-// If no latest height can be found after MaxRetryCount, it returns 0.
-func mustGetLatestHeight(ctx *parser.Context) uint64 {
```
guagualvcha, line 465: `mustGetLatestHeight`, to me the `must` means if it can not fetch the height in time, it should panic.

modular/blocksyncer/blocksyncer_options.go
```diff
@@ -1,478 +0,0 @@
...
-					// log.Debugw("enqueueing new block", "height", currHeight)
-					exportQueue <- currHeight
-				}
-			}
-		}
-	}
-}
-
-func (b *BlockSyncerModular) getLatestBlockHeight(ctx context.Context) {
-	for {
```
guagualvcha, line 210: I think we can use `time.NewTicker()` instead of `sleep` here. right now, if `LatestHeight` returns  error, it will keep doing this without any wait, thus cause high CPU as well.

modular/blocksyncer/blocksyncer_options.go
```diff
@@ -1,478 +0,0 @@
...
-				for ; currHeight <= uint64(latestBlockHeight); currHeight++ {
-					// log.Debugw("enqueueing new block", "height", currHeight)
-					exportQueue <- currHeight
-				}
-			}
-		}
-	}
-}
-
-func (b *BlockSyncerModular) getLatestBlockHeight(ctx context.Context) {
```
guagualvcha, line 209: `getLatestBlockHeight` is not a good function name, `syncBlockHeightLoop` seems better.

modular/blocksyncer/blocksyncer_options.go
```diff
@@ -1,478 +0,0 @@
...
-
-	// Create workers
-	worker := parser.NewWorker(b.parserCtx, exportQueue, 0, config.Cfg.Parser.ConcurrentSync)
-	worker.SetIndexer(b.parserCtx.Indexer)
-
-	latestBlockHeight := mustGetLatestHeight(b.parserCtx)
-	Cast(b.parserCtx.Indexer).GetLatestBlockHeight().Store(int64(latestBlockHeight))
-	go b.getLatestBlockHeight(ctx)
-
-	lastDbBlockHeight := uint64(0)
```
guagualvcha, line 165: To keep a same code style, suggest we wrap the logic here as a function, like `mustGetEpoch` with a retry number.

modular/blocksyncer/blocksyncer_options.go
```diff
@@ -1,478 +0,0 @@
...
-	cycle := uint64(0)
-	startBlock := uint64(0)
-	endBlock := uint64(0)
-	flag := 0
-
-	for {
-		latestBlockHeightAny := Cast(b.parserCtx.Indexer).GetLatestBlockHeight().Load()
-		latestBlockHeight := latestBlockHeightAny.(int64)
-		if latestBlockHeight == int64(endBlock) {
-			continue
```
guagualvcha, line 243: The logic here will also cause high CPU usage as well, please check.

modular/blocksyncer/blocksyncer_options.go
```diff
@@ -1,478 +0,0 @@
...
-		}
-		log.Info(count*(cycle+1) + startHeight - 1)
```
guagualvcha, line 245: We can not print logs in such a way, it is hard to read and understand.

modular/blocksyncer/blocksyncer_options.go
```diff
@@ -1,478 +0,0 @@
...
-		}
-	}
-}
-
-func (b *BlockSyncerModular) quickFetchBlockData(startHeight uint64) {
-	count := uint64(b.config.Parser.Workers)
-	cycle := uint64(0)
-	startBlock := uint64(0)
-	endBlock := uint64(0)
-	flag := 0
```
guagualvcha, line 237: `flag` is not a good variable name, and is it possible that we assign `endBlock` to be `startHeight-1`, so that we dont need this `flag` any longer.

modular/blocksyncer/blocksyncer_options.go
```diff
@@ -1,478 +0,0 @@
...
-				}
-				Cast(b.parserCtx.Indexer).GetLatestBlockHeight().Store(latestBlockHeight)
-
-				time.Sleep(time.Second)
-			}
-		}
-	}
-}
-
-func (b *BlockSyncerModular) quickFetchBlockData(startHeight uint64) {
```
guagualvcha, line 232: `quickFetchBlockData` --> `fetchBlockDataLoop`

modular/blocksyncer/blocksyncer_options.go
```diff
@@ -1,478 +0,0 @@
...
-	wg.Add(int(end - start + 1))
-	for i := start; i <= end; i++ {
-		go func(height uint64) {
-			defer wg.Done()
-
-			for {
-				block, err := b.parserCtx.Node.Block(int64(height))
-				if err != nil {
-					log.Warnf("failed to get block from node: %s", err)
-					continue
```
guagualvcha, line 288: Suggest to use ticker instead of `continue` directly.

modular/blocksyncer/blocksyncer_options.go
```diff
@@ -1,478 +0,0 @@
...
-				}
-
-				events, err := b.parserCtx.Node.BlockResults(int64(height))
-				if err != nil {
-					log.Warnf("failed to get block results from node: %s", err)
-					continue
-				}
-				txs, err := b.parserCtx.Node.Txs(block)
```
guagualvcha, line 296: I guess this maybe a critical problem here,  the `Node` here is a remote node, it means the underlying implementation is:
```
// Txs implements node.Node
func (cp *Node) Txs(block *tmctypes.ResultBlock) ([]*types.Tx, error) {
	txResponses := make([]*types.Tx, len(block.Block.Txs))
	for i, tmTx := range block.Block.Txs {
		txResponse, err := cp.Tx(fmt.Sprintf("%X", tmTx.Hash()))
		if err != nil {
			return nil, err
		}

		txResponses[i] = txResponse
	}

	return txResponses, nil
}
```

Will this start a lot of requests? Is it possible to turn it into a local process?

modular/blocksyncer/blocksyncer_options.go
```diff
@@ -1,478 +0,0 @@
...
-		break
-	}
-
-	// fetch block data
-	go b.quickFetchBlockData(lastDbBlockHeight + 1)
-
-	go b.enqueueNewBlocks(ctx, exportQueue, lastDbBlockHeight+1)
-
-	// Start each blocking worker in a go-routine where the worker consumes jobs
-	go worker.Start(ctx)
```
guagualvcha, line 182: the `worker.Start(ctx)` belongs to Juno , but few question here: 
a. `concurrentSync` of the `Worker` is only used to deal with error, does Workder really support concurrent sync here, only one worker as I can see.

modular/blocksyncer/blocksyncer_indexer.go
```diff
@@ -1,551 +0,0 @@
...
-// An error is returned if the operation fails.
-func (i *Impl) Processed(ctx context.Context, height uint64) (bool, error) {
-	ep, err := i.DB.GetEpoch(context.Background())
-	if err != nil {
-		return false, err
-	}
-	// log.Infof("epoch height:%d, cur height: %d", ep.BlockHeight, height)
-	if ep.BlockHeight > int64(height) {
-		heightKey := fmt.Sprintf("%s-%d", i.GetServiceName(), height)
-		blockMap.Delete(heightKey)
```
guagualvcha, line 515: This is tricky that delete item from this map when calling `Processed`, `Processed` is supposed to be a query function that changes nothing, please try not coupled the `Processed` with external logic.

modular/blocksyncer/blocksyncer_indexer.go
```diff
@@ -1,551 +0,0 @@
...
-	}
-	return lastBlockRecordHeight, err
-}
-
-func (i *Impl) GetLatestBlockHeight() *atomic.Value {
-	return &(i.LatestBlockHeight)
-
-}
-
-func (i *Impl) CreateMasterTable() error {
```
guagualvcha, line 544: Please delete this if it is no longer needed.

modular/blocksyncer/blocksyncer_indexer.go
```diff
@@ -1,551 +0,0 @@
...
-	currentEpoch, err := i.DB.GetEpoch(ctx)
-	if err == nil {
-		lastBlockRecordHeight = 0
-	} else {
-		lastBlockRecordHeight = uint64(currentEpoch.BlockHeight)
-	}
-	return lastBlockRecordHeight, err
-}
-
-func (i *Impl) GetLatestBlockHeight() *atomic.Value {
```
guagualvcha, line 539: why don't we return int64 directly, so the caller doesn't have to change by themselves.

And try to add a function `SetLatestBlockHeight()`  instead of `Cast(b.parserCtx.Indexer).GetLatestBlockHeight().Store(int64(latestBlockHeight))`

modular/blocksyncer/blocksyncer_indexer.go
```diff
@@ -1,551 +0,0 @@
...
-	log.Infof("height :%d tx count:%d event count:%d", height, txCount, eventCount)
-
-	blockMap.Delete(heightKey)
-	eventMap.Delete(heightKey)
-	txMap.Delete(heightKey)
-	i.ProcessedHeight = height
-
-	cost := time.Now().UnixMilli() - startTime
-	log.Infof("total cost: %d", cost)
-	metrics.BlocksyncerCatchTime.WithLabelValues(fmt.Sprintf("%d", height)).Set(float64(cost))
```
guagualvcha, line 171: I think this will cause the Prometheus metrics to become extreme large with thus dynamic labels.

modular/blocksyncer/blocksyncer_indexer.go
```diff
@@ -1,551 +0,0 @@
...
-			tmp = append(tmp, m.ObjectID)
-			bucketNameMap[m.BucketName] = tmp
-		case *models.StreamRecord:
-			streamRecords = append(streamRecords, m)
-		case *models.PaymentAccount:
-			paymentAccount = append(paymentAccount, m)
-		}
-	}
-
-	localDB := localdb.Cast(i.DB)
```
guagualvcha, line 324: why it is called `localDB`?

modular/blocksyncer/blocksyncer_indexer.go
```diff
@@ -1,551 +0,0 @@
...
-	for _, v := range resMap {
-		tmp := update[v.BucketName]
-		if tmp == nil {
-			tmp = make([]*models.Object, 0)
-		}
-		tmp = append(tmp, v)
-		update[v.BucketName] = tmp
-	}
-	if len(update) > 0 {
-		if err := localDB.BatchSaveObject(ctx, update); err != nil {
```
guagualvcha, line 386: If someone create an object and delete it at the same block, can block sync handle such case?

modular/blocksyncer/blocksyncer_indexer.go
```diff
@@ -1,551 +0,0 @@
...
-				objectEvent = append(objectEvent, e)
-			}
-			if prefixtree.BuildPrefixTreeEvents[event.Type] {
-				prefixEvent = append(prefixEvent, e)
-			}
-		}
-	}
-
-	if len(dataList) > 0 {
-		err := i.BatchHandle(ctx, dataList, block, common.Hash{})
```
guagualvcha, line 280: `BatchHandle` is a bad function name, as the following will invoke `concurrenceHandleEvent`, I did not get the difference between them from the function name.

modular/blocksyncer/blocksyncer_indexer.go
```diff
@@ -1,551 +0,0 @@
...
-	wg := &sync.WaitGroup{}
-	wg.Add(len(allEvents))
-	var handleErr error
-	for _, events := range allEvents {
-		go func(event []TxHashEvent) {
-			defer wg.Done()
-			for _, e := range event {
-				if err := i.HandleEvent(ctx, block, e.TxHash, e.Event); err != nil {
-					log.Errorf("failed to HandleEvent err:%v", err)
-					handleErr = err
```
guagualvcha, line 403: concurrent write to `handleErr`?

modular/blocksyncer/blocksyncer_indexer.go
```diff
@@ -1,551 +0,0 @@
...
-
-func (i *Impl) concurrenceHandleEvent(ctx context.Context, block *coretypes.ResultBlock, allEvents [][]TxHashEvent) error {
-	wg := &sync.WaitGroup{}
-	wg.Add(len(allEvents))
-	var handleErr error
-	for _, events := range allEvents {
-		go func(event []TxHashEvent) {
-			defer wg.Done()
-			for _, e := range event {
-				if err := i.HandleEvent(ctx, block, e.TxHash, e.Event); err != nil {
```
guagualvcha, line 401: I think there is some concurrent issue here: like an object is sealed first and then be deleted in a same block, the status of this object is not certain after the logic here.

modular/blocksyncer/blocksyncer_indexer.go
```diff
@@ -1,551 +0,0 @@
...
-
-// HandleMessage accepts the transaction and handles messages contained inside the transaction.
-func (i *Impl) HandleMessage(block *coretypes.ResultBlock, index int, msg sdk.Msg, tx *types.Tx) {
-	log.Info("HandleMessage")
-}
-
-// Processed tells whether the current Indexer has already processed the given height of Block
-// An error is returned if the operation fails.
-func (i *Impl) Processed(ctx context.Context, height uint64) (bool, error) {
-	ep, err := i.DB.GetEpoch(context.Background())
```
guagualvcha, line 508: Why not use the field  `ProcessedHeight` of `Impl`, get from db seems unnecessary.

modular/blocksyncer/blocksyncer_indexer.go
```diff
@@ -1,551 +0,0 @@
...
-
-	log.Infof("handle&write data cost: %d", time.Now().UnixMilli()-startTime)
-	log.Infof("height :%d tx count:%d event count:%d", height, txCount, eventCount)
-
-	blockMap.Delete(heightKey)
-	eventMap.Delete(heightKey)
-	txMap.Delete(heightKey)
-	i.ProcessedHeight = height
-
-	cost := time.Now().UnixMilli() - startTime
```
guagualvcha, line 169: Simplify it: `cost := time.Since(startTime)`

modular/blocksyncer/blocksyncer_options.go
```diff
@@ -1,478 +0,0 @@
...
-					log.Warnf("failed to get block results from node: %s", err)
-					continue
-				}
-				txs, err := b.parserCtx.Node.Txs(block)
-				if err != nil {
-					log.Warnf("failed to get block results from node: %s", err)
-					continue
-				}
-				heightKey := fmt.Sprintf("%s-%d", b.Name(), height)
-				blockMap.Store(heightKey, block)
```
guagualvcha, line 302: pay attention of the memory usage in case of the worker does not consume the data in time.

modular/blocksyncer/blocksyncer_options.go
constwz: Good advice. I'll change it

modular/blocksyncer/blocksyncer_options.go
constwz: Good advice. I'll change it

modular/blocksyncer/blocksyncer_options.go
constwz: Good advice. I'll change it

modular/blocksyncer/blocksyncer_options.go
constwz: This has been modified

modular/blocksyncer/blocksyncer_options.go
constwz: I'm going to delete this log

modular/blocksyncer/blocksyncer_options.go
constwz: Good advice. I'll change it

modular/blocksyncer/blocksyncer_options.go
constwz: I'll change my name and keep the code the same

modular/blocksyncer/blocksyncer_options.go
constwz: This has been modified

modular/blocksyncer/blocksyncer_options.go
constwz: Code version issue, this interface is no longer used

modular/blocksyncer/blocksyncer_options.go
constwz: There is something wrong with this comment, worker actually does not support concurrency, I delete it

modular/blocksyncer/blocksyncer_indexer.go
constwz: I'll delete this

modular/blocksyncer/blocksyncer_indexer.go
constwz: Good advice. I'll change it

modular/blocksyncer/blocksyncer_indexer.go
constwz: This has been fixed

modular/blocksyncer/blocksyncer_indexer.go
constwz: Code version issue. This is gone

modular/blocksyncer/blocksyncer_indexer.go
constwz: We guarantee the sequence of events in each module, the concurrency is just a different module concurrency, and this code is no longer available

modular/blocksyncer/blocksyncer_indexer.go
constwz: We guarantee the sequence of events in each module, the concurrency is just a different module concurrency, and this code is no longer available

modular/blocksyncer/blocksyncer_indexer.go
constwz: Version issue. This code has been removed

modular/blocksyncer/blocksyncer_indexer.go
constwz: We guarantee the sequence of events in each module, the concurrency is just a different module concurrency, and this code is no longer available

modular/blocksyncer/blocksyncer_indexer.go
constwz: We need to discuss that

PR #981, opened 2023-08-07
feat: add tx confirm func and create virtual group retry

will-2012:
### Description

Add tx confirm func and create virtual group retry.

### Rationale

Send tx and query tx hash, in order to ensure the tx executed by chain.

### Example

N/A.

### Changes

Notable changes: 
* signer/chain client.

### DIFF
--- a/base/gnfd/gnfd_service.go
+++ b/base/gnfd/gnfd_service.go
@@ -91,6 +91,11 @@ const (
 	ChainSuccessVerifyPutObjectPermission = "verify_putt_object_permission_success"
 	// ChainFailureVerifyPutObjectPermission defines the metrics label of unsuccessfully verify put object permission
 	ChainFailureVerifyPutObjectPermission = "verify_put_object_permission_failure"
+
+	// ConfirmBlockNumber defines wait block number.
+	ConfirmBlockNumber = 3
+	// WaitForNextBlockTimeout define wait block timeout.
+	WaitForNextBlockTimeout = 30 * time.Second
 )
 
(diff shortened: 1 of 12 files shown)


base/gnfd/gnfd_service.go
```diff
@@ -783,3 +787,70 @@ func (g *Gnfd) VerifyPutObjectPermission(ctx context.Context, account, bucket, o
 	}
 	return false, err
 }
+
+// ConfirmTransaction is used to confirm whether the transaction is on the chain.
+func (g *Gnfd) ConfirmTransaction(ctx context.Context, txHash string) (*sdk.TxResponse, error) {
```
Fynnss, line 792: It should have a timeout for retry.

will-2012: Currently, only retry once and there is a timeout in waitForNextBlock.

Fynnss (approved):

base/gnfd/gnfd_service.go
Fynnss: Why only retry once?

will-2012: Retrying once basically guarantees whether it is in the current block or the next block.
what is the suggestion here.

Fynnss: At least 3 blocks.

sysvm (approved):
LGTM

base/gnfd/gnfd_service.go
will-2012: Fixed.

PR #984, opened 2023-08-07
fix: when object not exist, don't return error for queryResumeOffsetHandler

jingjunLi:
### Description

remove this warning:
`"message":"do API error, url: http://127.0.0.1:9033/gsc/svm?upload-context=, err: statusCode 404 : code : 90003  (Message: no uploading record)"}`

### Rationale

### Example

NA

### Changes

Notable changes: 
* NA
* ...

### DIFF
--- a/modular/gater/object_handler.go
+++ b/modular/gater/object_handler.go
@@ -340,12 +340,10 @@ func (g *GateModular) queryResumeOffsetHandler(w http.ResponseWriter, r *http.Re
 	}
 
 	segmentCount, err = g.baseApp.GfSpClient().GetUploadObjectSegment(reqCtx.Context(), objectInfo.Id.Uint64())
-	if err != nil {
+	if err != nil && err.Error() != metadata.ErrNoRecord.String() {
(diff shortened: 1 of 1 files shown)


modular/gater/object_handler.go
```diff
@@ -341,7 +341,6 @@ func (g *GateModular) queryResumeOffsetHandler(w http.ResponseWriter, r *http.Re
 	segmentCount, err = g.baseApp.GfSpClient().GetUploadObjectSegment(reqCtx.Context(), objectInfo.Id.Uint64())
 	if err != nil {
```
will-2012, line 342: It is best to only ignore the not found error.

sysvm (approved):
LGTM

will-2012 (approved):

modular/gater/object_handler.go
will-2012: `if err != nil && err.Error() != metadata.ErrNoRecord.String() {
     return err
}` is more simple.

PR #987, opened 2023-08-08
fix: ignore duplicate entry when create bucket traffic

will-2012:
### Description

Ignore duplicate entry when create bucket traffic.

### Rationale

Ignore failures caused by concurrent updates.

### Example

N/A

### Changes

Notable changes: 
* downloader.


sysvm (approved):
LGTM

Fynnss (approved):

PR #988, opened 2023-08-08
fix: fix some bugs for bucket migrate

jingjunLi:
### Description

fix some bugs for bucket migrate

### Rationale

NA
### Example

NA

### Changes

Notable changes: 
* NA
* ...

### DIFF
--- a/modular/manager/bucket_migrate_scheduler.go
+++ b/modular/manager/bucket_migrate_scheduler.go
@@ -425,6 +425,7 @@ func (s *BucketMigrateScheduler) doneMigrateBucket(bucketID uint64) error {
 func (s *BucketMigrateScheduler) processEvents(migrateBucketEvents *types.ListMigrateBucketEvents) error {
 	// 1. process CancelEvents
 	if migrateBucketEvents.CancelEvents != nil {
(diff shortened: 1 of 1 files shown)


modular/manager/bucket_migrate_scheduler.go
```diff
@@ -459,18 +464,21 @@ func (s *BucketMigrateScheduler) processEvents(migrateBucketEvents *types.ListMi
 
 func (s *BucketMigrateScheduler) subscribeEvents() {
 	go func() {
-		UpdateBucketMigrateSubscribeProgressFunc := func() {
+		logNumber := uint64(0)
+
+		UpdateBucketMigrateSubscribeProgressFunc := func(idx uint64) {
```
will-2012, line 469: There is no need to pass parameters here, just use `logNumber` directly. And also `idx` is not a good parameter name.

will-2012 (approved):

modular/manager/bucket_migrate_scheduler.go
will-2012: The argument is a counter/number, not an index. idx represents the index normally.

Fynnss (approved):

PR #998, opened 2023-08-09
fix: fix bug for metadata crash (v0.2.4)

jingjunLi:
### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...


sysvm (approved):
LGTM

PR #1026, opened 2023-08-17
fix: report task when task finished

jingjunLi:
### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...


sysvm (approved):
LGTM

will-2012 (changes requested):
Please check executor AskTask func, which has reported progress.

PR #1067, opened 2023-08-28
feat: Docker image distroless update

dangerousidea:
### Description

To make the image itself & running env(Kubernetes) more secure,  this PR updates Dockerfile to use distroless base image.

### Rationale

To make the image itself & running env(Kubernetes) more secure. 

With this change, user will not be able to exec into the pod, as there's no shell, package manager and unused library. 

For distroless, please check [GoogleContainerTools/distroless](https://github.com/GoogleContainerTools/distroless)

### Changes

Notable changes: 
* Update Dockerfile base image and related statements;
* Update build.sh to use static linking
* Update builder image, replace build-base with gcc and libstdc++-dev to make it faster and smaller


XiaobinYuan (approved):

sysvm (approved):
LGTM

build.sh
```diff
@@ -14,6 +14,7 @@ fi
 buf generate
 
 go build -ldflags "\
+  -extldflags=-static
```
sysvm, line 17: Don't need static compiling.

dangerousidea: Added a `EXT_LD_FLAGS` env var in build.sh and set it `-static` only in distroless Dockerfile.

PR #1086, opened 2023-08-30
Feat bs ut

constwz:
### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...


modular/blocksyncer/test/mock_chian.go
```diff
@@ -0,0 +1,196 @@
+package test
```
sysvm, line 1: file name should be `mock_chain`?

sysvm:
UT commit msg should use `test: xxx`

ruojunm (approved):

BarryTong65 (approved):

PR #1092, opened 2023-09-01
fix: support reimburse quota if download consumed extra quota

flywukong:
### Description

fix quota consumed method. Use the actual downloaded size during download processing to calculate the value of the consumed quota,  if the extra quota has been consumed , the getObject handler will make a request to downloader to fix the quota traffic db

### Rationale

fix the download quota consumed method

### Example
NA

### Changes

Notable changes: 
* fix the download quota


store/sqldb/traffic.go
```diff
@@ -258,6 +258,91 @@ func (s *SpDBImpl) GetBucketTraffic(bucketID uint64, yearMonth string) (traffic
 	}, nil
 }
 
+// UpdateExtraQuota update the read consumed quota and free consumed quota in traffic db with the extra quota
+func (s *SpDBImpl) UpdateExtraQuota(bucketID, extraQuota uint64) error {
```
sysvm, line 262: Please add UT.

modular/downloader/download_task.go
```diff
@@ -279,17 +279,6 @@ func (d *DownloadModular) PreDownloadPiece(ctx context.Context, downloadPieceTas
 		bucketID := downloadPieceTask.GetBucketInfo().Id.Uint64()
 		bucketName := downloadPieceTask.GetBucketInfo().GetBucketName()
 
-		// if stream account has been frozen, it is not allowed to download
```
mliu233, line 282: Is the below logic no longer valid?

flywukong: no, if it is getObject handler , it is already checked in func (a *AuthenticationModular) VerifyAuthentication() . So This is a duplicate check

annielz (approved):

sysvm (approved):
LGTM

PR #1139, opened 2023-09-13
fix: refactoring code for getObject and Universal apis

ruojunm:
### Description

fix: refactoring code for getObject and Universal apis
### Rationale

1. When returning 406 in getObject, the http connection was closed and causing go-sdk can't parse the response
2. In Universal endpoint, it used "download Object" API, which can take up lots of memory.  
3. Many logic code are duplicates in both getObject and Universal apis. In this PR, they are refined and organized into private method and shared by both apis.

### Changes

Notable changes: 
* Remove unnecessary headers in error response
* Refactor  both getObject and Universal apis so that they can use the exact same download logic

### DIFF
--- a/modular/gater/object_handler.go
+++ b/modular/gater/object_handler.go
@@ -375,35 +375,14 @@ func (g *GateModular) queryResumeOffsetHandler(w http.ResponseWriter, r *http.Re
 // getObjectHandler handles the download object request.
 func (g *GateModular) getObjectHandler(w http.ResponseWriter, r *http.Request) {
 	var (
-		err                       error
-		reqCtxErr                 error
-		reqCtx                    *RequestContext
-		authenticated             bool
-		objectInfo                *storagetypes.ObjectInfo
-		bucketInfo                *storagetypes.BucketInfo
-		params                    *storagetypes.Params
-		lowOffset                 int64
-		highOffset                int64
-		pieceInfos                []*downloader.SegmentPieceInfo
-		pieceData                 []byte
-		extraQuota, consumedQuota uint64
-		replyDataSize             int
-		dbUpdateTimeStamp         int64
+		err           error
+		reqCtxErr     error
+		reqCtx        *RequestContext
+		authenticated bool
 	)
 	getObjectStartTime := time.Now()
 	defer func() {
 		if err != nil {
-			// if the bucket exists extra quota when download object, recoup the quota to user
-			if extraQuota > 0 {
-				quotaUpdateErr := g.baseApp.GfSpClient().RecoupQuota(reqCtx.Context(), bucketInfo.Id.Uint64(), extraQuota, sqldb.TimestampYearMonth(dbUpdateTimeStamp))
-				// no need to return the db error to user
-				if quotaUpdateErr != nil {
-					log.CtxErrorw(reqCtx.Context(), "failed to recoup extra quota to user", "error", err)
-				}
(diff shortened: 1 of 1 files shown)


modular/gater/object_handler.go
```diff
@@ -497,13 +476,55 @@ func (g *GateModular) getObjectHandler(w http.ResponseWriter, r *http.Request) {
...
+				// no need to return the db error to user
+				if quotaUpdateErr != nil {
+					log.CtxErrorw(reqCtx.Context(), "failed to recoup extra quota to user", "error", err)
+				}
+				log.CtxDebugw(reqCtx.Context(), "//"+
+					"success to recoup extra quota to user", "extra quota:", extraQuota)
+			}
+			w.Header().Del(ContentLengthHeader)
+			w.Header().Del(ContentRangeHeader)
+			w.Header().Del(ContentTypeHeader)
```
annielz, line 516: Why delete the ContentTypeHeader? I think we still need it to distinguish download or view?

modular/gater/object_handler.go
```diff
@@ -497,13 +476,55 @@ func (g *GateModular) getObjectHandler(w http.ResponseWriter, r *http.Request) {
...
+			w.Header().Del(ContentDispositionHeader)
```
annielz, line 517: ditto for ContentDispositionHeader

modular/gater/object_handler.go
ruojunm: Those deletion only happens when downloading failure happens. 

Now, in universal endpoint, setting ContentTypeHeader happens before the actual downloading process. 
So if the actual downloading failure, ContentTypeHeader needs to be cleaned.
Otherwise the error page html will be incorrectly displayed as object's content-type rather then text/html.

modular/gater/object_handler.go
ruojunm: same here.  If ContentDispositionHeader is not cleared, the error page html will be downloaded as a file rather than as a page.

Those slight differences are introduced because of the refactoring, which put "setting ContentDispositionHeader" and "setting ContentTypeHeader" in advance, so that the real actual download process could be reused in both getObjectHandler and universalEndpointHandler

annielz (approved):

modular/gater/object_handler.go
```diff
@@ -497,13 +476,55 @@ func (g *GateModular) getObjectHandler(w http.ResponseWriter, r *http.Request) {
...
 	} // else anonymous users can get public object.
 
+	// do the actual download
+	err = g.downloadObject(w, reqCtx)
+	if err != nil {
+		return
+	}
+}
+
+func (g *GateModular) downloadObject(w http.ResponseWriter, reqCtx *RequestContext) error {
```
flywukong, line 486: add comments for this fucntion .  And can we change the defination of function into  downloadObject(ctx context.Context , w http.ResponseWriter) ? context arg is usually put in the first place

flywukong (approved):

modular/gater/object_handler.go
clydemeng: comments are added. 
but context arg is not necessary as the this method will use the context of reqCtx :

```
reqCtx.Context()
```

flywukong: we can just pass reqCtx.Context() to downloadObject(ctx context.Context , w http.ResponseWrite), this may make it more general interface?

PR #1146, opened 2023-09-15
fix: fix itegrity hash check

alexgao001:
### Description

Currently, If uploaded object to SP does not match object meta created onchain, the SP does not handle the error and let replication to SSP happen, which should be restricted and also validation from secondary sp on receiving pieces must be performed to grantee consistence. 
 
### Rationale

Ensure data consistence
 
### Example

### Changes

Notable changes: 
* ...


guagualvcha:
Please and an integration test to demonstrate this is fixed.

guagualvcha (approved):

will-2012 (approved):

PR #1148, opened 2023-09-16
added feature for inputting OSS accesskeyid and accessSecretkey

atomstaker:
### Description
Hello，I am from Atomstaking
Recent days we tried to deployed one SP node on aliyun server, when we tried to start sp node with vairable accesskeyid and accessSecretkey in OS ENV, it is not so convinient:( and it is not so easy to set ENV variable if program is started in service mod on linux system.

so we think it is more convient to input these value at the start time if the program is stated in console
or use config file to set  accesskeyid and accessSecretkey for the program which is started in service mod on linux system

### Rationale
if this feature is added, now the program can be set variable value in 3 ways: console input, config file or enviroment variable.
it is more flexible and convenient

and I am just familiar with aliyun OSS , so I just changed code only for OSS
if you think this is reasonable and can be used for code, please add similiar function for other platform's piestore

Thanks

===============================================================
by the way we found there some wrong with the doc for starting SP node

in the doc :https://docs.bnbchain.org/greenfield-docs/docs/guide/storage-provider/run-book/piece-store

these seting for ENV enviroment is not correct:

PieceStore use environment variables of Alibaba Cloud OSS:

// OSSAccessKey defines env variable name for OSS access key
OSSAccessKey = "ALIBABA_CLOUD_ACCESS_KEY"
// OSSSecretKey defines env variable name for OSS secret key
OSSSecretKey = "ALIBABA_CLOUD_SECRET_KEY"
// OSSSessionToken defines env variable name for OSS session token
OSSSessionToken = "ALIBABA_CLOUD_SESSION_TOKEN"
// OSSRegion defines env variable name for OSS oss region
OSSRegion = "ALIBABA_CLOUD_OSS_REGION"

It should be like following according to your code:
	ALIBABA_CLOUD_ACCESS_KEY = "your aliyun accessKeyId"
	// OSSSecretKey defines env variable name for OSS secret key
	ALIBABA_CLOUD_SECRET_KEY = "your aliyun accessScretKey"
	// OSSSessionToken defines env variable name for OSS session token
	ALIBABA_CLOUD_SESSION_TOKEN =  "your aliyun session Key"
	// OSSRegion defines env variable name for OSS oss region
	ALIBABA_CLOUD_OSS_REGION = "your aliyun oss region"

please update your docs, otherwise other users will suffer same problem with me :(

thank

Atomstaker jia

### DIFF
--- a/cmd/storage_provider/main.go
+++ b/cmd/storage_provider/main.go
@@ -144,6 +144,13 @@ func storageProvider(ctx *cli.Context) error {
 		log.Errorw("failed to make gf-sp config", "error", err)
 		return nil
 	}
+	
+	err = checkAndInput(cfg)
+	if err != nil {
+		log.Errorw("failed to input keys", "error", err)
+		return nil
+	}	
+	
 	err = utils.MakeEnv(ctx, cfg)
 	if err != nil {
 		log.Errorw("failed to make gf-sp env", "error", err)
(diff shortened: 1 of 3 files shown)


sysvm:
Hello, thanks for your pr. Env environment in [PieceStore](https://docs.bnbchain.org/greenfield-docs/docs/guide/storage-provider/run-book/piece-store) is not described correctly. I'll fix it later.

PR #1155, opened 2023-09-19
fix: fix manager duplicate entry when inserting upload progress

sysvm:
### Description

1. Fix manager duplicate entry when inserting upload progress.
2. Fix query upload progress when record not found returns HTTP 404 not 500.

### Rationale

Optimize code logics.

### Example

N/A

### Changes

Notable changes: 
* N/A

### DIFF
--- a/modular/gater/object_handler.go
+++ b/modular/gater/object_handler.go
@@ -9,7 +9,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/bnb-chain/greenfield-storage-provider/store/sqldb"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
 	commonhttp "github.com/bnb-chain/greenfield-common/go/http"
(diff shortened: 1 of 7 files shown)


will-2012:
It is best to wait until the v0.2.6-alpha.2 version is stable before merging this PR.

modular/gater/object_handler.go
```diff
@@ -680,8 +680,12 @@ func (g *GateModular) queryUploadProgressHandler(w http.ResponseWriter, r *http.
 		taskState, errDescription, err = g.baseApp.GfSpClient().GetUploadObjectState(reqCtx.Context(), objectInfo.Id.Uint64())
 		if err != nil {
 			log.CtxErrorw(reqCtx.Context(), "failed to get uploading job state", "error", err)
+			if !strings.Contains(err.Error(), "no uploading record") {
```
will-2012, line 683: It is best to use some common functions for error handling such as non-existent/duplicate keys.

sysvm: If use a common function, there are many check conditions. I think write it in each function is ok.

will-2012 (approved):

guagualvcha (approved):

PR #1159, opened 2023-09-20
fix: list objects by gvg sql bug

BarryTong65:
### Description

fix list objects by gvg sql bug
To avoid SQL query length exceeding the server's max_allowed_packet size, we can add a simple check mechanism. One way is to keep track of the query string's length and break out of the loop if it gets too close to the limit.

### Rationale

```
client failed to list objects by gvg id, 
error: rpc error: code = Unknown desc = Error 1436 (HY000): 
Thread stack overrun: 242207 bytes used of a 262144 byte stack, and 20000 bytes needed.  Use &#39;mysqld --thread_stack=#&#39; to specify a bigger stack.
```

### Example

N/A

### Changes

Notable changes: 
* fix list objects by gvg sql bug

### DIFF
--- a/store/bsdb/object.go
+++ b/store/bsdb/object.go
@@ -404,7 +404,6 @@ func (b *BsDBImpl) ListObjectsInGVG(gvgID uint32, startAfter common.Hash, limit 
 		bucketIDsMap map[common.Hash]bool
 		tableNameMap map[common.Hash]string
 		gvgIDs       []uint32
-		query        string
 		err          error
 	)
 	startTime := time.Now()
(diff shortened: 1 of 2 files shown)


store/bsdb/object.go
```diff
@@ -443,16 +442,46 @@ func (b *BsDBImpl) ListObjectsInGVG(gvgID uint32, startAfter common.Hash, limit
...
-	return objects, buckets, err
+		// Append other groups in this chunk
+		for _, group := range chunk[1:] {
+			subQuery := fmt.Sprintf(" UNION ALL "+baseQuery, tableNameMap[group.BucketID], group.LocalVirtualGroupId, group.BucketID)
+			query += subQuery
+		}
+
+		// Apply the filter
+		query += fmt.Sprintf(filterQuery, startAfter)
+		finalQuery := fmt.Sprintf("select * from( %s", query)
```
sysvm, line 468: Why there are only one left parenthesis after from?

BarryTong65: image
there is a right parenthesis in the filterQuery.
here is the demo of the executed SQL
```
select * from( 
(select * from %s where local_virtual_group_id = %d and bucket_id = %s)
UNION ALL
(select * from %s where local_virtual_group_id = %d and bucket_id = %s)
) as combined where status = 'OBJECT_STATUS_SEALED' and object_id > %s and removed = false;
```

sysvm (approved):
LGTM

PR #1189, opened 2023-10-17
fix: add challenge v2 api to avoid piece hash http header too large

will-2012:
### Description

Add challenge v2 API to avoid piece hash HTTP header too large.

### Rationale

Components such as nginx can disconnect because the response header is too large, and users receive a 502 HTTP status code.

### Example

N/A.

### Changes

Notable changes: 
* Gateway add new challenge API v2.

### DIFF
--- a/modular/gater/admin_handler.go
+++ b/modular/gater/admin_handler.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/hex"
 	"encoding/json"
+	"encoding/xml"
 	"io"
 	"net/http"
 	"strings"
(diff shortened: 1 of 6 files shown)


modular/gater/admin_handler.go
```diff
@@ -377,6 +378,169 @@ func (g *GateModular) getChallengeInfoHandler(w http.ResponseWriter, r *http.Req
 	log.CtxDebug(reqCtx.Context(), "succeed to get challenge info")
 }
 
+// getChallengeInfoV2Handler handles get challenge piece info request. Current only greenfield
+// validator can challenge piece is store correctly. The challenge piece info includes:
```
sysvm, line 382: Currently, only greenfield validator can challenge piece whether is stored correctly.

modular/gater/admin_handler.go
```diff
@@ -377,6 +378,169 @@ func (g *GateModular) getChallengeInfoHandler(w http.ResponseWriter, r *http.Req
...
+// the challenged piece data, all piece hashes and the integrity hash. The challenger
+// can verify the info whether are correct by comparing with the greenfield info.
+// The difference between the v2 interface and the old interface is the format of the returned results,
```
sysvm, line 385: Maybe `the old interface(v1)` is better.

modular/gater/admin_handler.go
will-2012: Fixed.

modular/gater/admin_handler.go
will-2012: Fixed.

sysvm (approved):
LGTM

forcodedancing (approved):

PR #1203, opened 2023-10-25
fix: wrong error code  when invalid signature 

jingjunLi:
### Description

If there is a problem with the signature in the request, a 400 (ErrBadSignature) error should be returned to the user;

### Rationale

NA

### Example

NA

### Changes

Notable changes: 
* NA
* ...


sysvm (approved):
LGTM

will-2012 (approved):

PR #1207, opened 2023-10-26
feat: update recover object command for k8s job

annielz:
### Description

This pr update recovery object command to better fit k8s job with the following features:
1. Update the method to pass list of object into recovery command, using "//" to separate object names.
2. Add a check for current enqueued recovery jobs and let new jobs to wait until the queue is not too full.
3. Add current status of recovery queue and failed status into the stats api, for users to know their recovery progress and send a final report after finish.

### Rationale

For batch process recovery object k8s job and better user experience.

### Example

./gf-sp recover.object -config deployment/localup/local_env/sp1/config.toml -b sample-bucket -l objectName1//objectName2/objectName3

begin to recover the object of the primary SP: objectName1 
succeed to generate recovery object objectName1 task on background 
begin to recover the object of the primary SP: objectName2
succeed to generate recovery object d ef task on objectName2 
begin to recover the object of the primary SP: objectName3
succeed to generate recovery object gh task on objectName3 

Processing recovery tasks, update every 30 seconds. 
 Waiting Tasks Number: 3 
 Failed: []

Processing recovery tasks, update every 30 seconds. 
 Waiting Tasks Number: 2
 Failed: []

Finished process all recovery objects task on background. 
 Failed: [objectName2] 

### Changes

Notable changes: 
* manager


clydemeng (approved):

cmd/command/recovery.go
```diff
@@ -85,22 +89,46 @@ func recoverObjectAction(ctx *cli.Context) error {
...
 	if ctx.IsSet(objectListFlag.Name) {
-		if ctx.NArg() < 1 {
-			return fmt.Errorf("args number error")
-		}
-		argNum := ctx.NArg()
-		for i := 0; i < argNum-1; i++ {
-			recoverObject(bucketName, ctx.Args().Get(i), cfg, client)
+		objectNames := ctx.String(objectListFlag.Name)
+		objectNameList := strings.Split(objectNames, "//")
+		for _, objectName := range objectNameList {
```
flywukong, line 94: if len(objectNameList) == 0, return err

cmd/command/recovery.go
```diff
@@ -85,22 +89,46 @@ func recoverObjectAction(ctx *cli.Context) error {
...
 		objectName := ctx.String(objectFlag.Name)
 		recoverObject(bucketName, objectName, cfg, client)
 	}
+
+	taskCheckTicker := time.NewTicker(RecoveryJobSyncTime)
+
+	for range taskCheckTicker.C {
+		stats, _ := client.GetTasksStats(context.Background())
+		if stats.GetRecoveryProcessCount() > 0 {
+			log.Infof("Processing recovery tasks, update every 5 seconds. Waiting Tasks Number: %d. Recovered: %s. Failed: %s", stats.GetRecoveryProcessCount(), stats.GetRecoverySucceedList(), stats.GetRecoveryFailedList())
```
flywukong, line 107: If it is information for the user, it is recommended not to use log printing, but to directly use fmt.Println to print in separate lines. three lines for the three stat value

cmd/command/recovery.go
```diff
@@ -85,22 +89,46 @@ func recoverObjectAction(ctx *cli.Context) error {
...
 func recoverObject(bucketName string, objectName string, cfg *gfspconfig.GfSpConfig, client *gfspclient.GfSpClient) error {
-	var replicateIdx int
+	var (
+		replicateIdx    int
+		processingCount uint32
+	)
+
+	stats, _ := client.GetTasksStats(context.Background())
+	processingCount = stats.GetRecoveryProcessCount()
+	for processingCount >= MaxRecoveryJob {
```
flywukong, line 125: There may be multiple users and multiple threads calling the cmd recover command, and limiting the number of jobs called at a time may not have much effect.

cmd/command/recovery.go
annielz: If there is nothing after objectListFlag -l, the system will throw an error "flag needs an argument: -l". If there is anything after objectListFlag -l, objectNames will not be an empty string, and objectNameList := strings.Split(objectNames, "//") will never result in  len(bjectNameList) ==0 (e.g., strings.Split("example", "//") will result in [example]. Thus there is no need to add if len(objectNameList) == 0 check.

cmd/command/recovery.go
annielz: Got it, updated to fmt.Printf and use \n to separate lines. The output format is currently as below:
Processing recovery tasks, update every 5 seconds.
Waiting Tasks Number: 1
Recovered: [objectName1 objectName2]
Failed: []

cmd/command/recovery.go
annielz: The stats.GetRecoveryProcessCount() get the shared recoveryTaskMap from manager module,  and recoveryTaskMap will be shared across threads, so it shall give the real time jobs waiting for process.

leowkong (approved):

BarryTong98 (approved):

BarryTong65 (approved):

constwz (approved):

PR #1208, opened 2023-10-26
fix: verify object status in manager's reported replicate task

alexgao001:
### Description

A replicate task from manager queue might be asked by more than 1 executor. So that manager needs to check the object status from reported task and abort the it if object sealed. 

### Rationale

### Example

### Changes

Notable changes: 
* ...


modular/manager/manage_task.go
```diff
@@ -359,6 +359,17 @@ func (m *ManageModular) HandleReplicatePieceTask(ctx context.Context, task task.
 
 func (m *ManageModular) handleFailedReplicatePieceTask(ctx context.Context, handleTask task.ReplicatePieceTask) error {
 	if handleTask.GetNotAvailableSpIdx() != -1 {
+		objectInfo, queryErr := m.baseApp.Consensus().QueryObjectInfoByID(ctx, util.Uint64ToString(handleTask.GetObjectInfo().Id.Uint64()))
+		if queryErr != nil {
+			log.Errorw("failed to query object info", "object", handleTask.GetObjectInfo(), "error", queryErr)
+			return queryErr
+		}
+		if objectInfo.GetObjectStatus() == storagetypes.OBJECT_STATUS_SEALED {
+			log.CtxInfow(ctx, "object already sealed, abort replicate task")
```
will-2012, line 368: Print objectInfo.

will-2012 (approved):

sysvm (approved):
LGTM

modular/manager/manage_task.go
alexgao001: added

PR #1212, opened 2023-10-30
fix: return 400 when getting invalid header for getNonceAPI

ruojunm:
### Description

fix: return 400 when getting invalid header for getNonceAPI
### Rationale

N/A
### Example

 if user request doesn't contain a domain, it should response with a 400 httpCode rather than 500:
``` sh
curl --location '127.0.0.1:9133/auth/request_nonce' \
--header 'X-Gnfd-User-Address: 0x3d0a49B091ABF8940AD742c0139416cEB30CdEe0'

< HTTP/1.1 400 Bad Request
<Error><Code>50017</Code><Message>The X-Gnfd-App-Domain header is incorrect.</Message></Error>%
```

### Changes

Notable changes: 
* add domain/userAddress head parameter validation logic
* add UT

### DIFF
--- a/modular/gater/auth_handler.go
+++ b/modular/gater/auth_handler.go
@@ -10,6 +10,7 @@ import (
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/ethereum/go-ethereum/accounts"
+	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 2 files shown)


constwz (approved):
LGTM

BarryTong65 (approved):

annielz (approved):

PR #1214, opened 2023-10-31
fix: abort replication task if an object sealed

alexgao001:
### Description

A replicate task from manager queue might be asked by more than 1 executor. So that manager needs to check the object status from reported task and abort the it if object sealed.

### Rationale

### Example

### Changes

Notable changes: 
* ...


zjubfd (approved):

will-2012 (approved):

sysvm (commented):
Golang code lint action error can be ignored.

PR #1222, opened 2023-11-01
fix:bucket size write DB failed

constwz:
### Description

fix:bucket size write DB failed

### Rationale

bucket size Data fails to be written to the database due to large data volume. Therefore, data needs to be written in batches
### Example

nothing

### Changes

Notable changes: 
* Write bucket size data in batches


sysvm (approved):
LGTM

ruojunm (approved):

annielz (approved):

PR #1225, opened 2023-11-01
fix: sp should resume picking VGF regardless of RPC error

alexgao001:
### Description

the SP would check if its virtual group families are available by querying the chain before assigning VGF to a bucket. It should continue picking a VGF regardless of RPC error.  

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/base/gfspvgmgr/virtual_group_manager.go
+++ b/base/gfspvgmgr/virtual_group_manager.go
@@ -355,6 +355,8 @@ func (vgm *virtualGroupManager) PickVirtualGroupFamily() (*vgmgr.VirtualGroupFam
 	if err != nil {
 		return nil, err
 	}
+	vgm.mutex.RLock()
+	defer vgm.mutex.RUnlock()
 	return vgm.vgfManager.pickVirtualGroupFamily(filter)
 }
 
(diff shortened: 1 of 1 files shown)


sysvm (approved):
LGTM

base/gfspvgmgr/virtual_group_manager.go
```diff
@@ -349,8 +351,6 @@ func (vgm *virtualGroupManager) refreshMetaByChain() {
 // PickVirtualGroupFamily pick a virtual group family(If failed to pick,
 // new VGF will be automatically created on the chain) in get create bucket approval workflow.
 func (vgm *virtualGroupManager) PickVirtualGroupFamily() (*vgmgr.VirtualGroupFamilyMeta, error) {
-	vgm.mutex.RLock()
-	defer vgm.mutex.RUnlock()
```
will-2012, line 353: Why remove this lock guard.  vgfIDToVgf may be update/read concurrently in pickVirtualGroupFamily.

will-2012 (changes requested):

base/gfspvgmgr/virtual_group_manager.go
alexgao001: put it inside the `genVgfFilter`, because call to chain might take too long and it would lock the resource.

will-2012 (approved):

PR #1228, opened 2023-11-02
fix: primary sp should check received ssp signature before sealing

alexgao001:
### Description

primary sp should check received ssp bls signature before sealing object. If the bls signature is invalid. the secondary sp will be added to freeze pool as well.

### Rationale

### Example

### Changes

Notable changes: 
* ...


ruojunm (approved):

modular/executor/executor.go
```diff
@@ -83,16 +96,33 @@ func (e *ExecuteModular) eventLoop(ctx context.Context) {
 	}
 
 	statisticsTicker := time.NewTicker(time.Duration(e.statisticsOutputInterval) * time.Second)
+	updateSpTicker := time.NewTicker(3 * time.Second)
```
will-2012, line 99: Define const

will-2012 (approved):

PR #1230, opened 2023-11-02
fix: config sp black list

alexgao001:
### Description

1. SP in blacklist would not be picked for picking VGF.
2. GVG in blacklist would not be picked for creating GVG.

### Rationale

### Example

### Changes

Notable changes: 
* ...


will-2012 (changes requested):
This dev branch is wrong?

PR #1239, opened 2023-11-06
fix: add timeout for RPC calls to Secondary SP

alexgao001:
### Description

Set a timeout for calls to secondary SP, currently it is not config and the response from secondary SP might take over 2 min. This would block the executor to handle tasks. And if the Manager server can not receive a response from the executor, task will be dispatch which would cause duplicate task. 

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/modular/executor/execute_replicate.go
+++ b/modular/executor/execute_replicate.go
@@ -263,9 +263,9 @@ func (e *ExecuteModular) doReplicatePiece(ctx context.Context, waitGroup *sync.W
 		RtyDelay,
 		RtyErr,
 		retry.OnRetry(func(n uint, err error) {
-			log.CtxErrorw(ctx, "failed to replicate piece", "segment_idx", segmentIdx, "redundancy_idx", redundancyIdx, "error", err, "attempt", n, "max_attempts", RtyAttNum)
+			log.CtxErrorw(ctx, "failed to replicate piece", "sp_endpoint", spEndpoint, "segment_idx", segmentIdx, "redundancy_idx", redundancyIdx, "error", err, "attempt", n, "max_attempts", RtyAttNum)
 		})); err != nil {
-		log.CtxErrorw(ctx, "failed to replicate piece", "segment_idx", segmentIdx, "redundancy_idx", redundancyIdx, "error", err)
+		log.CtxErrorw(ctx, "failed to replicate piece", "sp_endpoint", spEndpoint, "segment_idx", segmentIdx, "redundancy_idx", redundancyIdx, "error", err)
 		return err
 	}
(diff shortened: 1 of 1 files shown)


modular/executor/execute_replicate.go
```diff
@@ -302,7 +313,9 @@ func (e *ExecuteModular) doneReplicatePiece(ctx context.Context, rTask coretask.
 	receive.SetSignature(taskSignature)
 	doneReplicateTime := time.Now()
 	if err = retry.Do(func() error {
```
will-2012, line 315: Why is there no callback parameter of retry.OnRetry?

will-2012 (approved):

modular/executor/execute_replicate.go
```diff
@@ -250,11 +253,19 @@ func (e *ExecuteModular) doReplicatePiece(ctx context.Context, waitGroup *sync.W
...
-			"redundancy_idx", redundancyIdx, "error", err)
+		// timeout for single piece replication
+		ctxWithTimeout, cancel := context.WithTimeout(ctx, replicateTimeOut)
+		defer cancel()
+		return e.baseApp.GfSpClient().ReplicatePieceToSecondary(ctxWithTimeout, spEndpoint, receive, data)
+	}, RtyAttem,
+		RtyDelay,
+		RtyErr,
+		retry.OnRetry(func(n uint, err error) {
+			log.CtxErrorw(ctx, "failed to replicate piece", "segment_idx", segmentIdx, "redundancy_idx", redundancyIdx, "error", err, "attempt", n, "max_attempts", RtyAttNum)
```
will-2012, line 266: It is best to print endpont and error in the context.

modular/executor/execute_replicate.go
```diff
@@ -250,11 +253,19 @@ func (e *ExecuteModular) doReplicatePiece(ctx context.Context, waitGroup *sync.W
...
+		})); err != nil {
+		log.CtxErrorw(ctx, "failed to replicate piece", "segment_idx", segmentIdx, "redundancy_idx", redundancyIdx, "error", err)
```
will-2012, line 268: Ditto, add the `spEndpoint`.

zjubfd (approved):

zjubfd (approved):

modular/executor/execute_replicate.go
alexgao001: added

modular/executor/execute_replicate.go
alexgao001: added

modular/executor/execute_replicate.go
alexgao001: added

PR #1240, opened 2023-11-06
feat: add a DataMigrationRecord table to save the process record for data migration tasks

constwz:
### Description

fix the bucket size logic

### Rationale

Fix the bucket size calculation logic

### Example

nothing

### Changes

Notable changes: 
* Modified the SQL


will-2012:
Refine the PR title.

PR #1245, opened 2023-11-13
fix: fix sp healthy checker dead lock bug

jingjunLi:
### Description

you can not have one thread call RLock twice and call RUnlock twice in sequence,and have an other thread call Lock at the same time.It will deadlock with little possibility.

reference: [sync: document that double RLock isn't safe #15418](https://github.com/golang/go/issues/15418)

### Rationale

should not use RLock twice in the same thread.

### Example

### Changes

Notable changes: 
* ...


will-2012 (changes requested):
Ignore.

will-2012 (approved):

sysvm (approved):
LGTM

PR #1252, opened 2023-11-24
fix:httpcode

constwz:
### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...

### Potential Impacts
* ...

### DIFF
--- a/base/types/gfsperrors/error.go
+++ b/base/types/gfsperrors/error.go
@@ -1,8 +1,10 @@
 package gfsperrors
 
 import (
+	"encoding/json"
 	"net/http"
 	"sort"
+	"strings"
 	"sync"
 )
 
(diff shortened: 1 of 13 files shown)


base/types/gfsperrors/error.go
```diff
@@ -43,6 +46,23 @@ func init() {
 	})
 }
 
+func GetGfSpErr(err error) *GfSpError {
```
ruojunm, line 49: what if the passed-in err is already a GfSpError?

constwz: Never mind. The.Error() method will also return a parsable string, which will then be parsed to form a GfSpError

base/types/gfsperrors/error.go
```diff
@@ -43,6 +46,26 @@ func init() {
...
+		return nil
+	}
+	if gfspErr, ok := err.(*GfSpError); ok {
+		return gfspErr
+	}
+	i := strings.Index(err.Error(), "desc = ")
+	if len(err.Error())-1 < i+6 {
+		return nil
+	}
+	errInfo := err.Error()[i+6:]
```
sysvm, line 60: Does `[i+6:]` fit all cases?

constwz: The result after the rpc call is "rpc error: code = Unknown desc = xxx" We only focus on xxx

PR #1253, opened 2023-11-24
fix: error code when object dose not exist

Bart200c:
### Description

fix error code when object dose not exist

### Rationale

### Example

### Changes

Notable changes: 
* ...

### Potential Impacts
* ...


sysvm (commented):
@BarryTong65 please help review.

BarryTong65:
GetObjectMeta is used to obtain the information of an object using the bucket name and object name. Additionally verifying the existence of the bucket is unnecessary.

PR #1257, opened 2023-11-30
fix:httpcode error

constwz:
Description
Addressing the issue of inaccurate error codes in the gateway interface.

Rationale
The error message is lost during the RPC call process, and we need to reparse it.

Example
“the specified bucket does not exist”
The error should return a 404, but it is actually returning a 500.

Changes
Notable changes:

Modified the string method of gfspError to make it easier to parse.
Added parsing logic to the gateway.
Potential Impacts
print log
http code

### DIFF
--- a/base/types/gfsperrors/error.go
+++ b/base/types/gfsperrors/error.go
@@ -46,26 +46,6 @@ func init() {
 	})
 }
 
-func GetGfSpErr(err error) *GfSpError {
-	if err == nil {
-		return nil
-	}
-	if gfspErr, ok := err.(*GfSpError); ok {
-		return gfspErr
-	}
-	i := strings.Index(err.Error(), "desc = ")
-	if i == -1 || len(err.Error())-1 < i+6 {
-		return nil
-	}
-	errInfo := err.Error()[i+6:]
-	var gfspErr GfSpError
-	unmarshalErr := json.Unmarshal([]byte(errInfo), &gfspErr)
-	if unmarshalErr == nil && gfspErr.HttpStatusCode != 0 {
-		return &gfspErr
-	}
-	return nil
-}
-
 // MakeGfSpError returns an GfSpError from the build-in error interface. It is
 // difficult to predefine all errors. For undefined errors, there needs to be a
 // way to capture them and return them to the client according to the GfSpError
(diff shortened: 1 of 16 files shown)


ruojunm (approved):

base/types/gfsperrors/error.go
```diff
@@ -56,20 +59,32 @@ func MakeGfSpError(err error) *GfSpError {
...
+	}
+	// Attempting to check if the error is an RPC error; if so, trying to extract its error description and convert it to the GfSpError type.
+	// Expected error should be: "rpc error: code = Unknown desc = xxx"
+	// We only need to focus on xxx, and can extract the substring using string.Index().
+	i := strings.Index(err.Error(), "desc = ")
+	if i != -1 && len(err.Error())-1 >= i+6 {
+		errInfo := err.Error()[i+6:]
+		var gfspErr GfSpError
+		unmarshalErr := json.Unmarshal([]byte(errInfo), &gfspErr)
+		if unmarshalErr == nil && gfspErr.HttpStatusCode != 0 {
```
will-2012, line 77: What does the 0 here mean?  200 means http is successful.

base/types/gfsperrors/error.go
```diff
@@ -56,20 +59,32 @@ func MakeGfSpError(err error) *GfSpError {
...
+		return gfspErr
+	}
+	// Attempting to check if the error is an RPC error; if so, trying to extract its error description and convert it to the GfSpError type.
+	// Expected error should be: "rpc error: code = Unknown desc = xxx"
+	// We only need to focus on xxx, and can extract the substring using string.Index().
+	i := strings.Index(err.Error(), "desc = ")
+	if i != -1 && len(err.Error())-1 >= i+6 {
+		errInfo := err.Error()[i+6:]
+		var gfspErr GfSpError
+		unmarshalErr := json.Unmarshal([]byte(errInfo), &gfspErr)
```
will-2012, line 76: Does the GRPC server use json serialization internally?

will-2012:
Refine the PR Title, `fix:httpcode error` --> `fix: httpcode error` to keep a consistent style, and maybe the title can be clearer.

base/types/gfsperrors/error.go
constwz: 0 indicates that err information is successfully resolved

base/types/gfsperrors/error.go
constwz: If you look at line 20, I've changed the encoding

constwz:
> Refine the PR Title, `fix:httpcode error` --> `fix: httpcode error` to keep a consistent style, and maybe the title can be clearer.

done

base/types/gfsperrors/error.go
```diff
@@ -56,20 +59,32 @@ func MakeGfSpError(err error) *GfSpError {
...
-			CodeSpace:      DefaultCodeSpace,
-			HttpStatusCode: int32(http.StatusInternalServerError),
-			InnerCode:      int32(DefaultInnerCode),
-			Description:    err.Error(),
+		return gfspErr
+	}
+	// Attempting to check if the error is an RPC error; if so, trying to extract its error description and convert it to the GfSpError type.
+	// Expected error should be: "rpc error: code = Unknown desc = xxx"
+	// We only need to focus on xxx, and can extract the substring using string.Index().
+	i := strings.Index(err.Error(), "desc = ")
```
sysvm, line 72: Define err.Error() as a variable.

base/types/gfsperrors/error.go
sysvm: Why not use `gfspErr.InnerCode != 0`? I think `gfspErr.HttpStatusCode` should be 200 by default.

sysvm (approved):
LGTM

alexgao001 (approved):

PR #1262, opened 2023-12-06
fix:  bug of the folder not being deleted

BarryTong65:
### Description

Fix the bug where deleting all objects within a folder at the same block height does not result in the folder being deleted

### Rationale

image

When a user deletes the objects beneath the /coco/ folder, leaving it empty, it should be deleted according to the block syncer logic. However, it is unexpectedly retained, resulting in a deleted folder "/coco/" appearing in the user's bucket.

```
"/coco/"
"/coco/data/train-00000-of-00040-67e35002d152155c.parquet"
"/coco/data/train-00004-of-00040-1df1755d6f7c9b17.parquet"
"/coco/data/train-00003-of-00040-a77e00648e423963.parquet"
"/coco/data/train-00001-of-00040-2c2b33b9504aa843.parquet"
"/coco/data/train-00002-of-00040-58e30306870b1de7.parquet"
```

image
https://greenfieldscan.com/bucket/0x00000000000000000000000000000000000000000000000000000000000007d8
When querying the user's bucket information on the browser side, all objects related to "/coco/" have disappeared, indicating that the bug resides here.
image

Root Cause:
The issue arises because subsequent optimizations in the block syncer aggregate all SQL queries within the same block height to execute together. Within a single block height, if I delete all objects under "/coco/", the deletion logic is processed, but the SQL is not executed yet. When it comes to deleting the last object, "/coco/data/train-00002-of-00040-58e30306870b1de7.parquet", the database still shows 5 objects under the "/coco/" folder (in reality, only the last one remains, and according to serial logic, "/coco/" should be deleted). As a result, both "/coco/" and "/coco/data/" folders are not deleted as they should be.

### Example

N/A

### Changes

Notable changes: 
* Execute all SQL queries in the prefix tree immediately, without proceeding to the step where they are concatenated into the final SQL statement, thus implementing the original code's serial logic to remove the bug.

### Potential Impacts
* QA Test @cd1129678576


constwz:
This change will cause the prefix table to wirte into the database first, and the object table has no data. Can you confirm that this situation has no impact on the query side?

modular/blocksyncer/database/prefixtree.go
```diff
@@ -10,21 +10,24 @@ import (
 )
 
 // CreatePrefixTree create prefix tree nodes by input slice
-func (db *DB) CreatePrefixTree(ctx context.Context, prefixTree []*bsdb.SlashPrefixTreeNode) (string, []interface{}) {
+func (db *DB) CreatePrefixTree(ctx context.Context, prefixTree []*bsdb.SlashPrefixTreeNode) error {
 	// because the passed in prefixTree is one object, the array members have same bucketName, we can use the first one
 	shardTableName := bsdb.GetPrefixesTableName(prefixTree[0].BucketName)
-	stat := db.Db.Session(&gorm.Session{DryRun: true}).Table(shardTableName).Create(&prefixTree).Statement
-	return stat.SQL.String(), stat.Vars
+	err := db.Db.WithContext(ctx).Table(shardTableName).Create(&prefixTree).Error
```
ruojunm, line 16: can you add e2e test cases in modular/blocksyncer/test/ ?

BarryTong98:
> This change will cause the prefix table to wirte into the database first, and the object table has no data. Can you confirm that this situation has no impact on the query side?

In 'prefix_tree.go', only the slash_prefix_tree_nodes table is queried, and it does not affect other logic.

PR #1280, opened 2023-12-18
chore: switch develop image to distroless

dangerousidea:
### Description

This PR switches the Dockerfile being used to `Dockerfile.distroless` for GitHub action 

### Rationale

To improve security within Kubernetes environment, it is recommended to use distroless image to prevent potential attack.

### Example

N/A

### Changes

Use `Dockerfile.distroless` to build. It has already been used in `signer`.

### Potential Impacts
The debugging will be harder, because after the change we cannot `exec into` the pods directly any more.


sysvm (approved):
LGTM

will-2012 (approved):

PR #1288, opened 2024-01-03
fix: replicate failed sp idx not included in local task

alexgao001:
### Description

The task reported from executor brings the notAvailableSPIdx if found replication failed. It should be used to check whether secondary SP is abnormal. 

### Rationale

### Example

### Changes

Notable changes: 
* ...

### Potential Impacts
* ...


will-2012 (approved):

will-2012 (commented):
Hotfix to master branch?

PR #1291, opened 2024-01-03
feat: add more log

jingjunLi:
### Description

When sp is unhealthy, we print more info.

### Rationale

### Example

### Changes

Notable changes: 
* ...

### Potential Impacts
* ...


alexgao001 (approved):

will-2012 (approved):
This is a hotfix to master branch??

PR #1305, opened 2024-01-15
fix: secondary SP might need to wait object meta due to latency

alexgao001:
### Description

1. Secondary SP might need to wait for object meta due to latency , so that it will not report err msg back to primary SP easily.
2. Shorten the SP freeze time to 5 min. 

### Rationale

### Example

### Changes

Notable changes: 
* ...

### Potential Impacts
* ...

### DIFF
--- a/modular/gater/admin_handler.go
+++ b/modular/gater/admin_handler.go
@@ -27,7 +27,11 @@ import (
 	sdktypes "github.com/cosmos/cosmos-sdk/types"
 )
 
-const MaxSpRequestExpiryAgeInSec int32 = 1000
+const (
+	MaxSpRequestExpiryAgeInSec int32 = 1000
+	checkPermissionRetry             = 3
+	checkPermissionSleepTime         = 3 * time.Second
+)
 
 // getApprovalHandler handles the get create bucket/object approval request.
 // Before create bucket/object to the greenfield, the user should the primary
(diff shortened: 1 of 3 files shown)


constwz (approved):

modular/gater/admin_handler.go
```diff
@@ -658,12 +662,22 @@ func (g *GateModular) replicateHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 func (g *GateModular) checkReplicatePermission(ctx context.Context, receiveTask gfsptask.GfSpReceivePieceTask, signatureAddr string) error {
-	objectInfo, err := g.baseApp.Consensus().QueryObjectInfo(ctx, receiveTask.ObjectInfo.BucketName, receiveTask.ObjectInfo.ObjectName)
+	var (
```
clydemeng, line 665: please add UT.

You can mock the first and second response for the same call

modular/gater/admin_handler.go
```diff
@@ -658,12 +662,22 @@ func (g *GateModular) replicateHandler(w http.ResponseWriter, r *http.Request) {
...
+		objectInfo *storagetypes.ObjectInfo
+		err        error
+	)
+	for retry := 0; retry < checkPermissionRetry; retry++ {
+		objectInfo, err = g.baseApp.Consensus().QueryObjectInfo(ctx, receiveTask.ObjectInfo.BucketName, receiveTask.ObjectInfo.ObjectName)
```
clydemeng, line 670: if we can't fetch objectInfo from chain, can we just retrieve it from SP metadata service?

alexgao001: It takes longer for meta  svc to have objects data stored.

modular/gater/admin_handler.go
alexgao001: added

ruojunm (approved):

PR #1333, opened 2024-01-29
fix: Not allow to upload piece/object for a fully-uploaded object

ruojunm:
### Description

Not allowed to upload piece or object for an object id which had already been fully uploaded before
### Rationale

Users might repeatedly upload pieces for a given object which had already been fully uploaded before. This might be due to some kinds of user code issues. But in sp side we should not allow this. 

Before this fix, sp will override the file content in piece store, which might cause "integrity check" not matching.

### Example

N/A
### Changes

Notable changes: 
* uploadTask: PreUploadObject
* uploadTask: PreResumableUploadObject

### Potential Impacts
N/A

### DIFF
--- a/modular/uploader/upload_task.go
+++ b/modular/uploader/upload_task.go
@@ -80,7 +80,8 @@ func (u *UploadModular) PreUploadObject(ctx context.Context, uploadObjectTask co
 		log.CtxErrorw(ctx, "failed to get upload object state")
 		return ErrGetObjectUploadState
 	}
-	if taskState == int32(types.TaskState_TASK_STATE_UPLOAD_OBJECT_DONE) {
+	if taskState == int32(types.TaskState_TASK_STATE_UPLOAD_OBJECT_DONE) ||
+		taskState == int32(types.TaskState_TASK_STATE_REPLICATE_OBJECT_DOING) {
 		// It is not allowed to upload piece or object for an object id which had already been fully uploaded.
 		log.CtxErrorw(ctx, "failed to put object as the target object had already fully uploaded")
 		return ErrInvalidUploadRequest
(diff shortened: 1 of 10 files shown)


modular/uploader/upload_task.go
```diff
@@ -312,6 +342,8 @@ func (u *UploadModular) HandleResumableUploadObjectTask(ctx context.Context, tas
 					ObjectID:  task.GetObjectInfo().Id.Uint64(),
 					TaskState: types.TaskState_TASK_STATE_UPLOAD_OBJECT_DONE,
 				})
+				log.CtxInfow(ctx, "object", task.GetObjectInfo(), "TaskState becomes types.TaskState_TASK_STATE_UPLOAD_OBJECT_DONE")
+				time.Sleep(time.Second * 10)
```
will-2012, line 346: Debugging code?

alexgao001, line 346: what does it mean?

modular/uploader/upload_task.go
```diff
@@ -70,6 +74,19 @@ func (u *UploadModular) PreUploadObject(ctx context.Context, uploadObjectTask co
...
+	if errGetUploadObjectState != nil {
+		log.CtxErrorw(ctx, "failed to get upload object state")
+		return ErrGetObjectUploadState
+	}
+	if taskState == int32(types.TaskState_TASK_STATE_UPLOAD_OBJECT_DONE) ||
+		taskState == int32(types.TaskState_TASK_STATE_REPLICATE_OBJECT_DOING) {
+		// It is not allowed to upload piece or object for an object id which had already been fully uploaded.
+		log.CtxErrorw(ctx, "failed to put object as the target object had already fully uploaded")
+		return ErrInvalidUploadRequest
+	}
```
will-2012, line 88: These codes can encapsulate a common check function, which is called by the upload and Resumable upload functions to facilitate unified modification; the state comparison does not seem to cover all and may cover more states.

modular/uploader/upload_task.go
clydemeng: fixed.

will-2012 (approved):

modular/uploader/upload_task.go
clydemeng: Fixed

alexgao001 (approved):

PR #1334, opened 2024-01-29
fix: gc object retry due to deleted bucket

BarryTong65:
### Description

The user uploads a very large object. Before the upload is completed, it is first partitioned by the primary SP and stored in OSS. During this process, if sealing is not successful (currently, a forced reject will delete it), and the user manually deletes it afterward, it is not being deleted from OSS.
image

Then I went to the blockchain to locate the corresponding object ID and found that it had not gone through the garbage collection (GC) process. Currently, there are two types of GC:

After a user executes events like cancel creation or deletion, a task is added to a queue in the background, which executes HandleGCObjectTask based on the block height.
HandleGCZombiePieceTask polls for deleted objects using their object ID. This hasn't happened yet for this particular object since it was only deleted a few days ago.

Currently, when querying the latest HandleGCObjectTask, it has only been executed up to block height 103102 and is continuously retrying due to execution failure.
image
image

image
image
The error occurs because when bucketInfo == nil, it triggers an error. In the process of executing the GC object and querying buckets, many deleted buckets are encountered, which then return nil, leading to continuous retries.

### Rationale

N/A

### Example

N/A

### Changes

Notable changes: 
* add get all bucket by name api to return all bucket info including deleted bucket.

### Potential Impacts
* QA


base/gfspclient/interface.go
```diff
@@ -147,6 +147,7 @@ type MetadataAPI interface {
 	PrimarySpIncomeDetails(ctx context.Context, spID uint32, opts ...grpc.DialOption) (int64, []*types.PrimarySpIncomeDetail, error)
 	SecondarySpIncomeDetails(ctx context.Context, spID uint32, opts ...grpc.DialOption) (int64, []*types.SecondarySpIncomeDetail, error)
 	GetBucketSize(ctx context.Context, bucketID uint64, opts ...grpc.DialOption) (string, error)
+	GetAllBucketByBucketName(ctx context.Context, bucketName string, opts ...grpc.DialOption) (*types.Bucket, error)
```
sysvm, line 150: The `GetAllBucketByBucketName` is confusing, it get all buckets but returned value is `*types.Bucket`, maybe `GetBucketInfoByBucketName` is better.

sysvm:
Don't add `clickup docs` in pr, it's a inner docs and SP is open source.

modular/executor/execute_gc.go
```diff
@@ -119,7 +119,7 @@ func (gc *GCWorker) isAllowGCCheck(objectInfo *storagetypes.ObjectInfo, bucketIn
 }
 
 func (gc *GCWorker) getGvgAndSpId(ctx context.Context, objectInfo *storagetypes.ObjectInfo) (*metadatatypes.Bucket, *virtualgrouptypes.GlobalVirtualGroup, uint32, error) {
-	bucketInfo, err := gc.e.baseApp.GfSpClient().GetBucketByBucketName(ctx, objectInfo.BucketName, true)
+	bucketInfo, err := gc.e.baseApp.GfSpClient().GetAllBucketByBucketName(ctx, objectInfo.BucketName)
 	if err != nil || bucketInfo == nil {
 		log.Errorw("failed to get bucket by bucket name", "bucket", bucketInfo, "error", err)
```
will-2012, line 124: Add bucket_name to the log context.

modular/executor/execute_gc.go
```diff
@@ -273,7 +273,7 @@ func (e *ExecuteModular) HandleGCObjectTask(ctx context.Context, task coretask.G
 			log.CtxDebugw(ctx, "delete the primary sp pieces", "object_info", objectInfo,
 				"piece_key", pieceKey, "error", deleteErr)
 		}
-		bucketInfo, err := e.baseApp.GfSpClient().GetBucketByBucketName(ctx, objectInfo.BucketName, true)
+		bucketInfo, err := e.baseApp.GfSpClient().GetAllBucketByBucketName(ctx, objectInfo.BucketName)
 		if err != nil || bucketInfo == nil {
 			log.Errorw("failed to get bucket by bucket name", "error", err)
```
will-2012, line 278: Ditto.

will-2012 (commented):
Please check the failed action.

base/gfspclient/interface.go
BarryTong65: fixed

BarryTong65:
> Don't add `clickup docs` in pr, it's an inner doc and SP is open source.

removed

modular/executor/execute_gc.go
BarryTong65: added

modular/executor/execute_gc.go
BarryTong65: added

will-2012 (approved):

base/gfspclient/metadata.go
```diff
@@ -1067,3 +1067,25 @@ func (s *GfSpClient) GetBucketSize(ctx context.Context, bucketID uint64, opts ..
...
+
+	req := &types.GfSpGetBucketInfoByBucketNameRequest{
+		BucketName: bucketName,
+	}
+
+	resp, err := types.NewGfSpMetadataServiceClient(conn).GfSpGetBucketInfoByBucketName(ctx, req)
+	ctx = log.Context(ctx, resp)
+	if err != nil {
+		log.CtxErrorw(ctx, "failed to send get all bucket rpc by bucket name", "error", err)
+		return nil, ErrRPCUnknownWithDetail("failed to send get all bucket rpc by bucket name, error: ", err)
```
sysvm, line 1088: Refine these logs.

modular/metadata/metadata_bucket_service.go
```diff
@@ -575,3 +575,56 @@ func (r *MetadataModular) GfSpGetBucketSize(ctx context.Context, req *types.GfSp
...
+		log.Errorw("failed to check bucket name", "bucket_name", req.BucketName, "error", err)
+		return nil, ErrInvalidBucketName
+	}
+
+	bucket, err = r.baseApp.GfBsDB().GetBucketInfoByBucketName(req.BucketName)
+	if err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return nil, ErrNoSuchBucket
+		}
+		log.CtxErrorw(ctx, "failed to get all bucket by bucket name", "error", err)
```
sysvm, line 597: ditto.

BarryTong65: fixed

base/gfspclient/metadata.go
BarryTong65: fixed

alanaladd21 (approved):

PR #1341, opened 2024-02-02
fix: let bucket migration be able to start without a pre-read of bucket content

ruojunm:
### Description

fix: let bucket migration be able to start without a pre-read of bucket content

### Rationale

Previously, the bucket migration could not be started as long as the bucket objects have never been read before, due to defect happens during GfSpGetLatestBucketReadQuota

### Example

N/A
### Changes

Notable changes: 
* Make it be able to return gorm.ErrRecordNotFound when reading traffic table if there is no corresponding record.
* Backfill the GfSpBucketQuotaInfo object when no record is found when reading traffic
* ut

### Potential Impacts
N/A

### DIFF
--- a/modular/metadata/metadata_bucket_service.go
+++ b/modular/metadata/metadata_bucket_service.go
@@ -417,13 +417,9 @@ func (r *MetadataModular) GfSpGetLatestBucketReadQuota(
 				freeQuotaSize = 0
 			}
 			quota := &gfsptask.GfSpBucketQuotaInfo{
-				BucketName:            bucketTraffic.BucketName,
-				BucketId:              bucketTraffic.BucketID,
-				Month:                 bucketTraffic.YearMonth,
-				ReadConsumedSize:      bucketTraffic.ReadConsumedSize,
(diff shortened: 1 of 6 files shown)


constwz (approved):

modular/metadata/metadata_bucket_service.go
```diff
@@ -417,13 +417,9 @@ func (r *MetadataModular) GfSpGetLatestBucketReadQuota(
 				freeQuotaSize = 0
 			}
 			quota := &gfsptask.GfSpBucketQuotaInfo{
-				BucketName:            bucketTraffic.BucketName,
```
mliu233, line 420: Is that we do not need to fill in BucketName, Month, ReadConsumedSize and FreeQuotaConsumedSize now for GfSpBucketQuotaInfo?

annielz (approved):

alexgao001 (approved):

modular/metadata/metadata_bucket_service.go
ruojunm: In this case, we don't.
the bucketTraffic is nil and it doesn't make sense to fill those values from nil

PR #1343, opened 2024-02-05
fix: correct the bucket quota formula when pre-check in bucket migration

ruojunm:
### Description

fix: correct the bucket quota formula when pre-check in bucket migration
### Rationale

The original comparison expression did not take the "ChargedQuotaSize" and "ReadConsumedSize" into account

### Example

N/A
### Changes
N/A

### Potential Impacts
N/A

### DIFF
--- a/modular/gater/migrate_handler.go
+++ b/modular/gater/migrate_handler.go
@@ -7,8 +7,9 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


will-2012 (approved):

flywukong (approved):
LGTM

PR #1352, opened 2024-03-01
feat: remove create bucket & object approval

BarryTong65:
### Description

Remove the create object approval & create bucket approval, and during the process of creating objects & buckets, the calls to the SP side are eliminated through the Go-sdk.

For CreateObject, validation for the object's existence has been added on chain, and the corresponding verify signature has been removed. 

Virtual Group Manager will automatically check the available status of each SP's GVG & VGF and create GVG when the storage is full or the SP does not have any available GVG.

#### Before 
SP Pick VGF & Create GVG process
image

#### Current
image
Replace the process on the SP side of checking for available VGF & GVG during the create bucket approval.
Virtual Group Manager will automatically check the available status of each SP's GVG & VGF and create GVG when the storage is full or the SP does not have any available GVG.

### Example

N/A

### Changes

Notable changes: 
* Virtual Group Manager will automatically check the available status of each SP's GVG & VGF and create GVG when the storage is full or the SP does not have any available GVG

### Potential Impacts
* Go-SDK

### DIFF
--- a/base/gfspconfig/config.go
+++ b/base/gfspconfig/config.go
@@ -222,8 +222,6 @@ type ParallelConfig struct {
 
 	LoadReplicateTimeout int64 `comment:"optional"`
 	LoadSealTimeout      int64 `comment:"optional"`
-
-	SyncAvailableVGFInterval int `comment:"optional"`
 }
 
 type TaskConfig struct {
--- a/base/gfspconfig/config.go
+++ b/base/gfspconfig/config.go
@@ -222,6 +222,8 @@ type ParallelConfig struct {
 
 	LoadReplicateTimeout int64 `comment:"optional"`
 	LoadSealTimeout      int64 `comment:"optional"`
+
+	SyncAvailableVGFInterval int `comment:"optional"`
 }
 
 type TaskConfig struct {
--- a/modular/manager/manager.go
+++ b/modular/manager/manager.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	sptypes "github.com/bnb-chain/greenfield/x/sp/types"
+	storagetypes "github.com/bnb-chain/greenfield/x/storage/types"
 	"golang.org/x/exp/slices"
 
 	"github.com/bnb-chain/greenfield-storage-provider/base/gfspapp"
@@ -26,7 +27,6 @@ import (
 	"github.com/bnb-chain/greenfield-storage-provider/pkg/metrics"
 	"github.com/bnb-chain/greenfield-storage-provider/store/types"
 	"github.com/bnb-chain/greenfield-storage-provider/util"
-	storagetypes "github.com/bnb-chain/greenfield/x/storage/types"
 )
 
 const (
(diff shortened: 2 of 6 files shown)


modular/manager/manager.go
```diff
@@ -1195,3 +1201,25 @@ func (m *ManageModular) QueryRecoverProcess(ctx context.Context, vgfID, gvgID ui
...
+		spList []*sptypes.StorageProvider
+	)
+
+	// query meta
+	if spList, err = m.baseApp.Consensus().ListSPs(context.Background()); err != nil {
+		log.CtxErrorw(ctx, "failed to list sps", "error", err)
+		return
+	}
+
+	for _, sp := range spList {
```
alexgao001, line 1217: no need to loop through the SP list

BarryTong65: Why is there no need to loop through the SP list? I need to ensure that every STATUS_IN_SERVICE SP has available VGF & GVG. If we don't query each SP in the SP lists to determine if a GVG needs to be created, it seems necessary to use a for loop to iterate through them.

BarryTong65: fixed

modular/manager/manager_options.go
```diff
@@ -87,6 +87,9 @@ const (
 	// on greenfield timeout height, if after current block height + timeout height, the object
 	// is not rejected, it is judged failed to reject unseal object on greenfield.
 	DefaultListenRejectUnSealTimeoutHeight int = 10
+	// DefaultSyncAvailableVGFInterval defines the default interval for available global virtual group family info,
+	// it is used to log and debug.
+	DefaultSyncAvailableVGFInterval int = 60
```
clydemeng, line 92: seconds, right?

modular/manager/manager.go
```diff
@@ -1195,3 +1201,24 @@ func (m *ManageModular) QueryRecoverProcess(ctx context.Context, vgfID, gvgID ui
...
+		sp  *sptypes.StorageProvider
+	)
+
+	// query meta
+	if sp, err = m.baseApp.Consensus().QuerySPByID(context.Background(), m.spID); err != nil {
+		log.CtxErrorw(ctx, "failed to list sps", "error", err)
+		return
+	}
+
+	// only pick vgf when sp is STATUS_IN_SERVICE
```
clydemeng, line 1217: do we need to take GVGPreferSPList and SPBlackList into account when picking vgf?

modular/manager/manager_options.go
BarryTong65: yes

modular/manager/manager.go
BarryTong65: SPBlackList will be used when picking VGF
image

GVGPreferSPList will be used when creating GVG
image

image

base/gfspconfig/config.go
```diff
@@ -222,6 +222,8 @@ type ParallelConfig struct {
 
 	LoadReplicateTimeout int64 `comment:"optional"`
 	LoadSealTimeout      int64 `comment:"optional"`
+
+	SyncAvailableVGFInterval int `comment:"optional"`
```
KeefeL, line 226: You can directly hardcode it in the code using macros.
Exporting as a config will increase the cognitive and maintenance costs of the SP.

BarryTong65: fixed with default value 5 mins

ruojunm (approved):
LGTM

KeefeL (approved):

PR #1364, opened 2024-03-18
fix: fix issues

alexgao001:
### Description

1. The gorm delete func not working as expect, use the query instead. 
2. Only delegated upload need to persist secondary piece hash into DB 
3.  The `objectSize` is missing in shadowObejctMeta
4. the `GCZombieByintegrityMeta` should only reply on Blocksyner data. if object is deleted, chain wont have the meta so that not able to delete piece from piece store since version info is missing.

### Rationale

### Example

### Changes

Notable changes: 
* ...

### Potential Impacts
* ...


base/types/gfsptask/gc.go
```diff
@@ -313,12 +313,15 @@ func (m *GfSpGCStaleVersionObjectTask) InitGCStaleVersionObjectTask(priority cor
 	redundancyIndex int32,
 	integrityChecksum []byte,
 	pieceChecksumList [][]byte,
-	version, timeout int64) {
+	version int64,
+	objectSize uint64,
```
constwz, line 317: objectSize What does this field do in the gc task？

alexgao001: we need to copy the integrate meta from the shadow one. ObjectSize is needed

ruojunm (approved):
LGTM

constwz (approved):

PR #1435, opened 2024-11-14
feat: fix gc objects

annielz:
### Description

Currently in sp, if an object is deleted within the same tx as its local virtual group (lvg), the async GC Task will have issue finding its secondary sps, as this info is only available by using logic objectId -> lvg -> gvg -> secondary sps.

In this case, it used to be that the data in secondary sp's data storage will not be removed. Now updating logic for all sps to check for files that have prefix related to the object (as "e{objectid}_"), and remove all related files.

### Rationale

Update logic of GC 

### Example
https://testnet.greenfieldscan.com/tx/0x2a34610146b15f60728e79b96181dc2af60825578822f70b913348b6cb8bed28
This link is an example of object & lvg in the same tx.

### Changes

Notable changes: 
* GC Task
* new data storage methods

### Potential Impacts
* GC Task

### DIFF
--- a/modular/executor/execute_gc.go
+++ b/modular/executor/execute_gc.go
@@ -201,7 +201,6 @@ func (e *ExecuteModular) HandleGCObjectTask(ctx context.Context, task coretask.G
 		currentGCBlockID   uint64
 		currentGCObjectID  uint64
 		responseEndBlockID uint64
-		storageParams      *storagetypes.Params
 		gcObjectNumber     int
 		tryAgainLater      bool
 		taskIsCanceled     bool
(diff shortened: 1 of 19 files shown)


constwz (approved):
/LGTM

modular/executor/execute_gc.go
```diff
@@ -288,21 +290,33 @@ func (e *ExecuteModular) HandleGCObjectTask(ctx context.Context, task coretask.G
...
+			for rIdx, sspId := range gvg.GetSecondarySpIds() {
+				if spId == sspId {
+					redundancyIndex = int32(rIdx)
+					for segIdx := uint32(0); segIdx < segmentCount; segIdx++ {
+						pieceKey := e.baseApp.PieceOp().ECPieceKey(currentGCObjectID, segIdx, uint32(rIdx), objectVersion)
+						if objectInfo.GetRedundancyType() == storagetypes.REDUNDANCY_REPLICA_TYPE {
+							pieceKey = e.baseApp.PieceOp().SegmentPieceKey(objectInfo.Id.Uint64(), segIdx, objectVersion)
+						}
+						// ignore this delete api error, TODO: refine gc workflow by enrich metadata index.
+						deleteErr := e.baseApp.PieceStore().DeletePiece(ctx, pieceKey)
```
clydemeng, line 303: can we we "DeletePiecesByPrefix" in this case as well?

mliu233: Done. Updated both primary sp & secondary sp's delete logic to batch delete pieces with object prefix.

clydemeng (approved):

ruojunm (approved):

